PR #107, opened 2018-02-14
History/history view for ceg

mikmakmuk:
Try out by adding and deleting some elements from a model. The view needs some polishing as I'm not sure how the history is going to be used. Currently its a bit noisy (even though I filtered out element feature changes where the values remained the same). Suggestions welcome.

Check out also the FIXME in `HistoryProviderImpl`. Not sure if this should/can be fixed.

### DIFF
--- a/bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/HistoryProviderImpl.java
+++ b/bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/HistoryProviderImpl.java
@@ -1,6 +1,7 @@
 package com.specmate.persistency.cdo.internal;
 
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
@@ -10,7 +11,10 @@ import org.eclipse.emf.cdo.CDOObjectHistory;
 import org.eclipse.emf.cdo.common.commit.CDOChangeSetData;
 import org.eclipse.emf.cdo.common.commit.CDOCommitInfo;
 import org.eclipse.emf.cdo.common.id.CDOID;
+import org.eclipse.emf.common.util.ECollections;
+import org.eclipse.emf.common.util.TreeIterator;
 import org.eclipse.emf.ecore.EObject;
+import org.eclipse.emf.ecore.EReference;
 import org.eclipse.emf.ecore.EStructuralFeature;
 import org.osgi.service.component.annotations.Component;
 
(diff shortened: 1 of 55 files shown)


bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/HistoryProviderImpl.java
```diff
@@ -79,9 +118,27 @@ protected void changedObject(CDOID id, EStructuralFeature feature, EChangeKind c
...
+			}
+			
 			Change change = HistoryFactory.eINSTANCE.createChange();
 			change.setFeature(feature.getName());
-			change.setValue(newValue.toString());
+			 
+			if(newValue != null)
+				change.setNewValue(newValue.toString());
+			else {
+				//FIXME when elements are deleted, the deltaprocessor does not receive them as detached objects but rather as
```
sebeder: If possible, avoid FIXME comments. In my experience, these  things are never fixed in the end. Or is there a good reason for this here?

web/src/app/modules/views/controller/modules/view-controller/services/view-controller.service.ts
```diff
@@ -46,6 +48,10 @@ export class ViewControllerService {
         return this.selectedElementService.hasSelection;
     }
 
+    public get historyShown(): boolean {
+        return Type.is(this.navigator.currentElement, CEGModel);
```
sebeder: Do we really provide history only for CEGModels? I think we should include several other classes here, like Process, TestSpecification, and TestProcedure. I know, the branch name is a little differnt to that , but we should do that, because it is just done then.

sebeder: Or, we show the history for all classes (and skip this check by returning true all the time). Are there elements, where we do not want to display the history? @junkerm says this is correct - display it all the time.

bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/HistoryProviderImpl.java
mikmakmuk: I added this to get attention in the review. It worked :). 
My idea was that FIXMEs are removed before merged to develop.

web/src/app/modules/views/controller/modules/view-controller/services/view-controller.service.ts
mikmakmuk: Ok, will fix.

web/src/app/modules/views/side/modules/history-view/components/history-view.component.html
```diff
@@ -0,0 +1,18 @@
...
+  <ul class="list-group list-group-flush">
+    <li *ngFor="let entry of modelHistoryEntries" class="list-group-item">
+      <span class="data-item-title">{{entry.user}}</span>&nbsp;<small class="text-muted">{{entry.date.toLocaleString()}}</small>
+        <ul>
+            <li *ngFor="let change of entry.changes" class="list-unstyled">
+              <div *ngIf="change.isDelete">
+                {{'ElementDeleted' | translate}}
+              </div>
+              <div *ngIf="!(change.isCreate && change.isDelete)">
+                {{'Changed' | translate}} <span class="data-item-title">{{change.feature}}</span>&nbsp;<span class="fa fa-arrow-right" aria-hidden="true">&nbsp;{{change.newValue}}</span>
```
sebeder: The span with the icon leads to ill-rendered font in the text behind the icon. Place the icon in a span- or i-tag and close it immediately.

PR #110, opened 2018-02-20
Uc7/bug in search

junkerm:
- Fix bug when searching for "something -"
- fixed specmate-integration-test project
- created test-cases for search


junkerm:
I changed the search in the UI project explorer to search for terms that have the query as prefix

PR #111, opened 2018-02-21
Handling the back/forward buttons of browsers in navigation

mikmakmuk:
Two main changes:
1) Use always the subscription on the routing events to handle data retrieval,
not only when constructing the navigator service. In this way, we load
also data when the browser back/forward buttons are used for navigation.
2) In order to keep the navigator service history in sync when the
browser back/forward buttons are used, we update the history index.


web/src/app/modules/navigation/modules/navigator/services/navigator.service.ts
```diff
@@ -18,20 +19,26 @@ export class NavigatorService {
         private dataService: SpecmateDataService,
         private logger: LoggingService,
         private router: Router,
-        private route: ActivatedRoute) {
+        private route: ActivatedRoute,
+        private location: Location) {
+
+        this.location.subscribe(pse => {
```
sebeder: Would it be possible to handle all of this stuff with the location, and just omit the ActivatedRoute? I never really understood this anyways.

mikmakmuk: I'm not sure we can. 
We use route to get the currentUrl, which looks for example as follows:
`test-data/Folder-1/Requirement-4`
From location, we can get the path, which looks for example as follows:
`/-/Requirement/test-data%2FFolder-1%2FRequirement-4`

I can urldecode it, but then we still have the basePath which is constructed in `url.ts`. I can implement there a function stripBasePath(path) which would return the url, needed for the dataservice to return the correct element. Would that work?

sebeder: Yes, that should work.

junkerm:
I merged the develop branch into this branch. Now I get a warning "Element could not be read" when I start specmate. The reason seems to be an invalid access of "/services/rest/details" on startup (maybe the project explorer). I don't get this behavior on the develop branch, so it seems it has to do with the changes in this branch.

PR #114, opened 2018-03-16
Fix persistency configuration logic

mikmakmuk:
The persistency service did not consider the configuration provided
by the configuration service at all. Now, we inject the configuration
admin and get relevant configuration items from there and fall only back
to hardcoded default values if the configuration in the file is not complete.

### DIFF
--- a/bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/CDOPersistencyService.java
+++ b/bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/CDOPersistencyService.java
@@ -1,7 +1,9 @@
 package com.specmate.persistency.cdo.internal;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Dictionary;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/CDOPersistencyService.java
```diff
@@ -281,19 +284,39 @@ public IView openView() {
...
 	}
 
-	private boolean readConfig(Config config) {
-		jdbcConnection = config.cdoJDBCConnection();
-		repository = config.cdoRepositoryName();
-		resourceName = config.cdoResourceName();
-		userResourceName = config.cdoUserResourceName();
-		if (StringUtils.isEmpty(repository) || StringUtils.isEmpty(resourceName)
-				|| StringUtils.isEmpty(userResourceName)) {
+	private boolean readConfig(Config defaultConfig) {
```
junkerm: the config that is provided here is not the default config but the config from the configuration admin. In fact, the configuration admin calls the activate method with the saved config.

bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/CDOPersistencyService.java
```diff
@@ -281,19 +284,39 @@ public IView openView() {
...
+		try {
+			Dictionary<String, Object> properties = configurationAdmin.getConfiguration(CDOPersistenceConfig.PID).getProperties();
```
junkerm: this is not necessary, the properties obtained here are the same as the parameter provided in the activate method (or should be, otherwise there is an error somewhere else)

mikmakmuk:
PR based on wrong assumptions. Ignore.

PR #115, opened 2018-03-17
Migration

mikmakmuk:
Migration of db when one attribute is added.


bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -0,0 +1,229 @@
...
+import com.specmate.migration.test.support.ServiceController;
+import com.specmate.model.support.util.SpecmateEcoreUtil;
+import com.specmate.persistency.IPackageProvider;
+import com.specmate.persistency.IPersistencyService;
+import com.specmate.persistency.ITransaction;
+import com.specmate.persistency.cdo.internal.CDOPersistencyService;
+import com.specmate.persistency.cdo.internal.config.CDOPersistenceConfig;
+import com.specmate.urihandler.IURIFactory;
+
+public class AddAttributeTest {
```
junkerm: General remark: To test addition of an attribute, maybe using the id attribute is not the best idea, as several other services rely on this attribute being present. In fact executing the test triggers several exceptions that are written on the console. My suggestion would be to use a new attribute.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -0,0 +1,229 @@
...
+	}
+	
+	@Test
+	public void doMigration() throws Exception {
+		activatePersistency(baselineModelController.getService());
+		ITransaction transaction = persistencyService.openTransaction();
+		Resource resource = transaction.getResource();
+		EObject root = SpecmateEcoreUtil.getEObjectWithName("root", resource.getContents());
+		assertNotNull(root);
+		EObject diagram = SpecmateEcoreUtil.getEObjectWithName("d0", root.eContents());
```
junkerm: maybe add assertTrue(root instanceof com.specmate.migration.test.baseline.testmodel.base.Folder) to be sure

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -0,0 +1,229 @@
...
+		assertNotNull(diagram);
+		assertNull(SpecmateEcoreUtil.getAttributeValue(root, "id", String.class));
```
junkerm: maybe add assertTrue(root instanceof com.specmate.migration.test.baseline.testmodel.artefact.Diagram) to be sure

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -0,0 +1,229 @@
...
+	public void testNeedsMigration() throws Exception {
+		activatePersistency(baselineModelController.getService());
+		assertFalse(migratorService.needsMigration());
+		migratorService.setModelProviderService(attributeAddedModelController.getService());
+		assertTrue(migratorService.needsMigration());
+		deactivatePersistency();
+	}
+	
+	@Test
+	public void doMigration() throws Exception {
```
junkerm: for readability reasons you could break this method up in several methods

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -0,0 +1,229 @@
...
+		assertNotNull(root);
+		EObject diagram = SpecmateEcoreUtil.getEObjectWithName("d0", root.eContents());
+		assertNotNull(diagram);
+		assertNull(SpecmateEcoreUtil.getAttributeValue(root, "id", String.class));
+		assertNull(SpecmateEcoreUtil.getAttributeValue(diagram, "id", String.class));
+		transaction.close();
+		deactivatePersistency();
+		
+		migratorService.setModelProviderService(attributeAddedModelController.getService());
+		migratorService.doMigration();
```
junkerm: why do you need to tirgger this manually? Should'nt the persistency service detect the need for migration and trigger the migration itself?

bundles/specmate-migration/src/com/specmate/migration/basemigrators/AttributeAddedBaseMigrator.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+import com.specmate.common.SpecmateException;
+
+public abstract class AttributeAddedBaseMigrator {
+	protected Connection connection;
+
```
junkerm: A more outspoken name for the methods would be migrateNewStringAttribute

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: I'm not sure I'm following. Do you refer to the fact that the ecore model objects have an attribute "id"? The tests on this level are intended to be agnostic of the business logic, so the attribute could be called "foo" or "bar", for the tested functionality (adding an attribute) it does not matter.
It is due to the lack of my imagination that the ecore model has these object and attributes. I can rename them and model, for example, food, to make clear that the models that are tested have nothing in common with the models used in specmate.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: Yep, makes sense.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: Yes, make sense.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: Yes, agree.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
junkerm: yes, the tests are agnostic about the attribute but, e.g. the persistency service is not. For instance it generates events when something changes, and for these events it uses the id attribute. Hence, when you run the current tests, you will find errros on the console "could not determine uri from model" or something like this. This is due to the missing id attribute.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: This is due to the the following:
1) I need to control the life-cycle of the CDOPersistencyService in order to be able to (a) generate the database and data, and (b) run the migration in ONE execution of the test cases. I must be able to replace the data model at runtime and this is only possible if I can shut down the restart the persistency service.
2) When I start the persistency service manually (see ServiceController), the CDOPersistencyService.activate() method is NOT called automatically by the OSGi framework. The activate method of the persistency service performs the migration, which I cannot call since it expects a "Config" object which is private to the CDOPersistencyService class. Hence, as a workaround, I have added an activate method that I can call from the test. This is ugly, so if there is another way to solve this, without changing business code, I'm open to suggestions.

junkerm: ok, so then let's leave like this for now (maybe you could add a code comment)

bundles/specmate-migration/src/com/specmate/migration/basemigrators/AttributeAddedBaseMigrator.java
mikmakmuk: Yes, makes sense.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: Ah, I see. Ok.

bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
mikmakmuk: There is already a comment in CDOPersistencyService that explains the purpose of the added activate method. I've added there what we need to consider when fixing this workaround.

junkerm: sorry for being picky :-) I ment a comment in the test procedure (line 120)

mikmakmuk: ;) done

PR #118, opened 2018-03-23
Migration: add several attributes

mikmakmuk:
Main changes:
1) Test a model with 3 new attributes of different type
2) refactoring of migration test code in a common base class
3) Instead of implementing one test model provider for each model, we use now only one implementation that decides, based on a configuration option retrieved by ConfigurationAdmin, which model should be provided
4) Instead of implementing one test migrator for each test case, we use now only one implementation that performs different migrations, depending on the executed test case. We use the same strategy as in 3).

Change 3) led to the removal of code that was added earlier to the production code, just to be able to execute the migration tests.
Change 4) is necessary to even be able to test different migrators.


junkerm:
I noticed that in CDOPersistencyService the activiate method for testing is just named "activate". Please rename it to activateFromTest or similar.

junkerm:
One question: I didn't see at which place you fixed the issue with more than one attribute. Was the problem merely in the test?

mikmakmuk:
Turned out that the problem is not multiple attributes, but attributes of primitive type. New test case is on the way. I did not add it here as this PR was already large enough with all the refactorings.

PR #121, opened 2018-03-30
Maintenance mode

junkerm:
1) Adds a service to enable and disable maintenance mode
2) When in maintenance mode modifying rest calls are blocked


bundles/specmate-administration/src/com/specmate/administration/api/ESpecmateStatus.java
```diff
@@ -0,0 +1,25 @@
+package com.specmate.administration.api;
+
+public enum ESpecmateStatus {
+
+	NORMAL(ESpecmateStatus.NORMAL_NAME, false), MAINTENANCE("maintenance", true);
+
```
mikmakmuk: Use MAINTENANCE_NAME constant instead of literal

bundles/specmate-emfrest-api/src/com/specmate/emfrest/api/RestServiceBase.java
```diff
@@ -99,6 +99,11 @@ public Object delete(Object object) throws SpecmateException, SpecmateValidation
 		return null;
 	}
 
+	@Override
+	public boolean isStatusService() {
+		return false;
+	}
+
```
mikmakmuk: Hm, wouldn't it be more elegant if StatusRestService would implement IStatusService, instead of using composition? Because then you would not need to have this isStatusService() method but can check via instanceof if something is a IStatusService (which is exported in the API).

bundles/specmate-integration-test/src/com/specmate/test/integration/AdministrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+	private void enterMode(String mode) {
+		JSONObject status = getStatusObject(mode);
+		String statusUrl = buildUrl("status");
+		RestResult<JSONObject> result = restClient.post(statusUrl, status);
+		Assert.assertEquals(Status.OK.getStatusCode(), result.getResponse().getStatus());
+	}
+
+	private void enterMaintenanceMode() {
+		enterMode("maintenance");
+	}
```
mikmakmuk: Use constants defined in API

bundles/specmate-integration-test/src/com/specmate/test/integration/AdministrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+	private void enterNormalMode() {
+		enterMode("normal");
+	}
```
mikmakmuk: Use constants defined in API

bundles/specmate-integration-test/src/com/specmate/test/integration/AdministrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+	private void checkIsInMode(String mode) {
+		String url = buildUrl("status");
+		JSONObject status = getObjectByUrl(Status.OK.getStatusCode(), url);
+		Assert.assertEquals(status.get(AdministrationPackage.Literals.STATUS__VALUE.getName()), mode);
+	}
+
+	private void checkIsInMaintenanceMode() {
+		checkIsInMode("maintenance");
+	}
+
```
mikmakmuk: Use constants defined in API

bundles/specmate-integration-test/src/com/specmate/test/integration/AdministrationTest.java
```diff
@@ -0,0 +1,88 @@
...
+	private void checkIsInNormalMode() {
+		checkIsInMode("normal");
+	}
```
mikmakmuk: Use constants defined in API

bundles/specmate-administration/src/com/specmate/administration/api/ESpecmateStatus.java
junkerm: done

bundles/specmate-emfrest-api/src/com/specmate/emfrest/api/RestServiceBase.java
junkerm: done

PR #123, opened 2018-04-03
Robust commit

junkerm:
Implement a mechanism that repeats a change to the repository and commits until a limit is reached or the commit was successfull


bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/TransactionImpl.java
```diff
@@ -59,13 +70,39 @@ public void commit() throws SpecmateException {
...
+	}
+
+	@Override
+	public <T> T doAndCommit(IChange<T> change) throws SpecmateException, SpecmateValidationException {
+		int maxAttempts = 10;
+		boolean success = false;
+		int attempts = 1;
+		T result = null;
+
+		while (!success && attempts <= maxAttempts) {
```
mikmakmuk: Hm, does this not cause additional problems if there is no (short) wait between attempts? If timing is an issue, as I understood from your explanation last meeting, maybe adding an increasing delay (50 ms after 1st failure, 100ms after second, 200ms after third etc) would be better than hammering the request 10 times out, as fast as possible...

junkerm: good comment --> fixed

PR #129, opened 2018-04-10
Validation/additional process model checks

sebeder:
See: https://trello.com/c/OS1Mlieg/157-process-model-start-mit-eingehender-oder-mehreren-ausgehenden-verbindungen-ende-mit-ausgehender-verbindung-aktivit%C3%A4t-mit-mehrere

### DIFF
--- a/web/src/app/config/config.ts
+++ b/web/src/app/config/config.ts
@@ -105,5 +105,5 @@ export class Config {
     public static ERROR_PROCESS_END_OUTGOING_CONNECTION = 'End node with outgoing connection.';
     public static ERROR_PROCESS_START_INCOMING_CONNECTION = 'Start node with incoming connection.';
     public static ERROR_PROCESS_NODE_MULTIPLE_OUTGOING_CONNECTIONS = 'Non-decision-node with multiple outgoing connections.';
-    public static ERROR_PROCESS_DECISION_WITH_ONE_OR_LESS_OUTGOING_CONNECTIONS = 'Decision node with too less outgoing connections.';
+    public static ERROR_PROCESS_DECISION_WITH_ONE_OR_LESS_OUTGOING_CONNECTIONS = 'Decision node not at least two outgoing connections.';
 }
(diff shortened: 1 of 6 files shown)


web/src/app/config/config.ts
```diff
@@ -102,4 +102,8 @@ export class Config {
     public static ERROR_MISSING_CONDITION = 'Missing condition.';
     public static ERROR_MISSING_FIELDS = 'Missing fields: {{fields}}';
     public static ERROR_NO_STEPS = 'No steps in model.';
+    public static ERROR_PROCESS_END_OUTGOING_CONNECTION = 'End node with outgoing connection.';
+    public static ERROR_PROCESS_START_INCOMING_CONNECTION = 'Start node with incoming connection.';
+    public static ERROR_PROCESS_NODE_MULTIPLE_OUTGOING_CONNECTIONS = 'Non-decision-node with multiple outgoing connections.';
+    public static ERROR_PROCESS_DECISION_WITH_ONE_OR_LESS_OUTGOING_CONNECTIONS = 'Decision node with too less outgoing connections.';
```
junkerm: too less --> not at least two

web/src/app/validation/process/decision-multiple-outgoing-connections-validator.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ProcessDecision } from '../../model/ProcessDecision';
+import { ProcessConnection } from '../../model/ProcessConnection';
+
+@Validator(Process)
+export class DecisionMultipleOutgoingConnectionsValidator extends ElementValidatorBase<Process> {
+    public validate(element: Process, contents: IContainer[]): ValidationResult {
+        const invalidNodes: IContainer[] = contents
+            .filter((element: IContainer) => Type.is(element, ProcessDecision))
+            .filter((element: ProcessDecision) =>
+                element.outgoingConnections && element.outgoingConnections.length < 2);
```
junkerm: what is if element.outgoingConnection == undefined? In this case there would be no outgoing connections (which means the decision node is invalid) BUT the validator would not flag the decision node as invalid.

junkerm (changes requested):

web/src/app/validation/process/decision-multiple-outgoing-connections-validator.ts
sebeder: This should be covered by a different check: node-no-outgoing-validator.ts. There, we flag all node, that do not have outgoind nodes as invalid. (Of course, not end nodes)

junkerm (approved):

PR #137, opened 2018-04-17
Displaying error markers at all graphical elements

sebeder:
See https://trello.com/c/zplaQ5gm/156-process-model-bei-start-ende-und-decision-fehlt-das-rote-ausrufezeichen-im-symbol-wenn-das-feld-name-leer-ist

We wanted to have error markers at all nodes and edges in the edited graphs.

### DIFF
--- a/web/src/app/modules/views/main/editors/modules/graphical-editor/elements/graphical-element-base.ts
+++ b/web/src/app/modules/views/main/editors/modules/graphical-editor/elements/graphical-element-base.ts
@@ -18,6 +18,7 @@ export abstract class GraphicalElementBase<T extends IContainer> {
     }
 
     public get isValid(): boolean {
-        return this.validationService.isValid(this.element);
+        return this.validationService.isValid(this.element) &&
+            this.validationService.currentInvalidElements.find((element: IContainer) => element === this.element) === undefined;
     }
 }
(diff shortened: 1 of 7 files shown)


web/src/app/modules/views/main/editors/modules/graphical-editor/elements/graphical-element-base.ts
```diff
@@ -18,6 +18,7 @@ export abstract class GraphicalElementBase<T extends IContainer> {
     }
 
     public get isValid(): boolean {
-        return this.validationService.isValid(this.element);
+        return this.validationService.isValid(this.element) &&
+            this.validationService.currentInvalidElements.find((element: IContainer) => element === this.element) === undefined;
```
junkerm: probably there is good reason, but I don't get why is there an isValid method AND the currentInvalidElements collection? This seems redundant.

junkerm (changes requested):

junkerm (commented):
The warning icons for start and end nodes are a bit hard to see. If possible without much effort it would be nicer to put them outside the black circle

web/src/app/modules/views/main/editors/modules/graphical-editor/elements/graphical-element-base.ts
sebeder: Yes, there is: we perform the isValid check not on each element seperately for performance reasons. We rather check, e.g., the model and its contents, and not all its children seperately, because for validating the model, we have to check its children anyways. I moved the check, however.

PR #140, opened 2018-04-19
Migration documentation

mikmakmuk:


bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
```diff
@@ -84,33 +87,44 @@ private void closeConnection() throws SpecmateException {
 
 	@Override
 	public boolean needsMigration() throws SpecmateException {
```
junkerm: in case of an exception or if you are unable to initiate the db connection I wouldn't return false, but instead throw an exception to signal that something unexpected happened. Handling the exception is then the job of the client.

bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
```diff
@@ -87,8 +87,10 @@ private void closeConnection() throws SpecmateException {
 
 	@Override
 	public boolean needsMigration() throws SpecmateException {
-		if (!initiateDBConnection()) {
-			logService.log(LogService.LOG_WARNING, "Migration: Could not obtain connection. Does the database exist?");
+		try {
+			initiateDBConnection();
+		} catch (SpecmateException e) {
+			logService.log(LogService.LOG_WARNING, e.getMessage());
 			return false;
```
junkerm: also here i would rather rethrow the exception to make clear that something bad happened. If we return false, the client will assume all is fine and no migration is needed, when in fact a migration might be needed but we just cannot determine it. Probably we should't even start Specmate if we cannot detmine if a migration is needed.

mikmakmuk: I've simplified the logic and treat everything as an error (that should be handled by the client), except the case when the database does not exist. See source code in last commit for an explanation.

bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/CDOPersistencyService.java
```diff
@@ -147,14 +147,7 @@ public void modified(Config config) throws SpecmateException {
...
 		if (migrationService.needsMigration()) {
-			logService.log(LogService.LOG_INFO, "Data migration needed.");
-			if (migrationService.doMigration()) {
-				// Successful migration
-				logService.log(LogService.LOG_INFO, "Data migration successful.");
-			} else {
-				logService.log(LogService.LOG_ERROR, "Data migration failed.");
-				throw new SpecmateException("Data migration failed.");
-			}
+			migrationService.doMigration();
```
junkerm: what if the migration fails, will the persistency be started? It probably shouldn't be started right?

PR #142, opened 2018-04-20
Migration improved CDO

mikmakmuk:
Removal of test specific code in CDOPersistency service and refactoring of test cases

### DIFF
--- a/bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
+++ b/bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
@@ -63,7 +63,6 @@ public class MigratorService implements IMigratorService {
 			String jdbcConnection = (String) properties.get(CDOPersistenceConfig.KEY_JDBC_CONNECTION);
 			//jdbcConnection = "jdbc:h2:./database/specmate";
 			this.connection = DriverManager.getConnection(jdbcConnection + ";IFEXISTS=TRUE", "", "");
-			System.out.println(connection.toString());
 		} catch (SQLException e) {
 			System.out.println(e.getMessage());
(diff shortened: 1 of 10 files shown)


bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
```diff
@@ -92,6 +92,7 @@ public boolean needsMigration() throws SpecmateException {
 			// TODO using the exception for logic decisions is not good as we
 			// have more than 1 reason
 			// why an exception can occur.
+			System.out.println(e.getMessage());
```
junkerm: remove System.out.println, optionally replace with LogService call

bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
```diff
@@ -92,6 +92,7 @@ public boolean needsMigration() throws SpecmateException {
 			// TODO using the exception for logic decisions is not good as we
```
junkerm: should we act on this todo now? Otherwise it will probably never happen.

bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
mikmakmuk: Should be fixed in PR 140.

bundles/specmate-migration/src/com/specmate/migration/internal/services/MigratorService.java
mikmakmuk: Should be fixed in PR 140.

PR #143, opened 2018-04-24
Making side windows resizable

sebeder:
See card: https://trello.com/c/7tSfWn9q/154-rechte-spalte-properties-change-history-traces-links-actions-die-einzelnen-bl%C3%B6cke-sollten-ein-ausklappbar-sein


munterkalmsteiner (approved):
Manually tested and works ok.

PR #146, opened 2018-05-02
Back-end implementation for authentication

mikmakmuk:
Base infrastructure that allows login/logout and checks requests whether
they belong to an authenticated session.


bundles/specmate-auth/src/com/specmate/auth/internal/UserSession.java
```diff
@@ -0,0 +1,29 @@
+package com.specmate.auth.internal;
+
+import java.util.Date;
+
+public class UserSession {
+	private static final long MAXIMUM_AGE = 60 * 60 * 12 * 1000L; // 12 hours in milliseconds
```
sebeder: Im future, we might want to make this configurable via a config-file. I think we should make this configurable right away.

bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/Login.java
```diff
@@ -0,0 +1,60 @@
...
+	}
+	
+	@Override
+	public boolean canGet(Object target) {
+		return true;
+	}
+	
+	@Override
+	public Object get(Object object, MultivaluedMap<String, String> queryParams) {
+		String username = queryParams.getFirst(USERNAME_PARAM);
```
sebeder: I'm just thinking: Is it right to send username and password? For now, the traffic between client and server is not encrypted. We definitely hav to change this as soon as we deploy this.

bundles/specmate-emfrest/src/com/specmate/emfrest/internal/AuthentificationFilter.java
```diff
@@ -1,70 +1,75 @@
...
+			return;
+		}
+		
+		// Get the Authorization header from the request
+        String authorizationHeader =
+                requestContext.getHeaderString(HttpHeaders.WWW_AUTHENTICATE);
+
+        // Validate the Authorization header
+        if (authorizationHeader == null) {
+        	logService.log(LogService.LOG_INFO, "Invalid authorization header: " + authorizationHeader);
```
sebeder: For debugging reasons: Could you add, how the header should actually look like?

bundles/specmate-auth/src/com/specmate/auth/internal/AuthentificationServiceImpl.java
```diff
@@ -1,92 +1,51 @@
...
-				logService.log(LogService.LOG_INFO, "Attempt to authenticate with wrong password: " + username);
-				return false;
-			}
-		}
+	public String authenticate(String username, String password, String projectname) throws SpecmateException {
+		/*
+		 * TODO retrieve the access rights for this user and set in the user session.
+		 */
+		
+		String token = randomString.nextString();
```
sebeder: Am I right, that for now, every user/pass-combination is accepted? I need this for testing.

bundles/specmate-auth/src/com/specmate/auth/internal/UserSession.java
mikmakmuk: Ok, will fix.

bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/Login.java
mikmakmuk: Using SSL, the query parameters are encrypted. Having said that, confidential data can still leak, e.g. in web-server logs. Hence, I agree that sending plain text passwords, even over SSL, is not a good idea.
I will improve the implementation and try to follow the design of [Digest access authentication](https://en.wikipedia.org/wiki/Digest_access_authentication).

bundles/specmate-emfrest/src/com/specmate/emfrest/internal/AuthentificationFilter.java
mikmakmuk: Currently, it is simple: WWW-Authenticate: token
This will change with the Digest access authentication, mentioned above.

bundles/specmate-auth/src/com/specmate/auth/internal/AuthentificationServiceImpl.java
mikmakmuk: Yes.

bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/Login.java
sebeder: The problem with this will be that we need username and password (plain) to authenticate against the hp-tools.

mikmakmuk: Hm, allright, then I understand. Ok, then we MUST use SSL. To diminish the chance of data leakage to log files, we send username/password in the http header instead of the URL. Not optimal, but best I can think of now.

sebeder: Yeah, this is not optimal, but the only solution for now, I guess. We could encrypt u/p with a key exchanged using Diffie-Hellman. Is that overkill?

mikmakmuk: Sorry, what I meant was to make a POST request, instead of sending the credentials in the URL.

mikmakmuk: Yes, asymmetric encryption would work. The client uses the public key to encrypt the data. The server decrypts it with the private key. We don't need key exchanges, AFAICS.

sebeder: Assymetric encryption would only work, if we do not store the private key in the repository. I was thinking about encrypting symetrically, but using a secret key exchanged using Diffie-Hellman. But I think the assymetric version is better, because it is just easier. We have to make the keys, however, configurable.

mikmakmuk: I think we should choose a protocol as secure (but not less) as the other systems we communicate with. We can discuss this through other channels :)

bundles/specmate-emfrest/src/com/specmate/emfrest/internal/AuthentificationFilter.java
mikmakmuk: Sorry, made a mistake here. The correct header is: 
`Authorization: token`
I fixed it in commit 20e55341639b22c9be06a80521d4f4e3c27bad22.

mikmakmuk:
This PR has been merged with PR 149, hence closing.

PR #149, opened 2018-05-08
Login screen

sebeder:
As we just agreed: We accept this merge (if there are no major comprohensibility issues) and accept that there are bugs. We will address this bugs in seperate issues to be able to develop further.


munterkalmsteiner (approved):
Front-end code looks good. Nothing that particularly stands out as strange or difficult to understand.

PR #150, opened 2018-05-11
Authentication/logout user on invalidated token

sebeder:


web/src/app/modules/views/main/authentication/modules/auth/services/authentication.service.ts
```diff
@@ -9,22 +9,40 @@ import { IContainer } from '../../../../../../../model/IContainer';
...
+        }
+        return UserToken.INVALID;
+    }
+    public set token(token: UserToken) {
+        this.cookie.put(
+            AuthenticationService.TOKEN_COOKIE_KEY,
+            JSON.stringify(token),
+            {expires: new Date(new Date().getTime() + 24 * 60 * 60 * 1000)});
+    }
+
```
mikmakmuk: Hm, how is the cookie expiration time linked with the maximum session idle time on the server?
1) Should the cookie expiration be refreshed after each request? I did not see that happen in the code, but that may not mean anything...
2) Should the cookie expiration time be the same as the session maximum idle time? In the back-end, this variable is configurable and stored in the specmate-config.properties. I guess this cannot be used when building the front-end?

munterkalmsteiner (commented):
Please see comment. If I'm not making sense, merge ;)

web/src/app/modules/views/main/authentication/modules/auth/services/authentication.service.ts
sebeder: You're absolutely right with these points. To 1.: We should refresh that - I'll change that. 2.: I do not see a point in doing this, since the user will be logged out, as soon as the token expires in the backend. However, we need to delete the cookie then. This way, the cookie expiration date does not really matter anymore, but it should be longer than the expiration date on the server. This way, the backend invalidates the cookie indirectly. And it should not matter, if an old token exeists in a cookie of the client. I'll change that.

munterkalmsteiner (approved):
Makes sense. So we have no expiration time on the client at all, right?

PR #151, opened 2018-05-11
Project configuration

junkerm:
Enables configuring projects with connectors and exporters. 
Connectors can authenticate user credentials


munterkalmsteiner (approved):
Looks good to me (except for the integration-test failures).

munterkalmsteiner:
I have now found the root cause of the problem. The project service needs the config service to startup. The authentication service, in turn, needs the project service. The integration tests need the authentication service. However, the integration tests need to be configured manually, and when we startup the config service (by proxy through authentication and project service), we loose that control. 

I've solved that by implementing a dummy project service in the integration test bundle that doesn't pull in the config service.

junkerm:
I refactored the way projects and connectors and exporters are wired together. Projects are now separate services that can be created independent from the configuration (e.g. for the dummy data).

bundles/specmate-integration-test/src/com/specmate/test/integration/support/DummyProjectService.java
```diff
@@ -13,8 +15,28 @@
...
+			public IRequirementsSource getConnector() {
+				// TODO Auto-generated method stub
+				return null;
+			}
+
+			@Override
+			public IExportService getExporter() {
+				// TODO Auto-generated method stub
+				return null;
+			}
```
munterkalmsteiner: Remove TODOs

munterkalmsteiner (approved):
Just a minor comment. Good solution! Ready to merge.

PR #152, opened 2018-05-12
Migration dateattribute

munterkalmsteiner:


bundles/specmate-migration-test/src/com/specmate/migration/test/AddAttributeTest.java
```diff
@@ -39,6 +41,8 @@ protected void checkMigrationPostconditions() throws Exception {
 		Diagram d0 = (Diagram) diagram;
 		assertNull(d0.getName());
 		d0.setName("d0");
+		Date date = d0.getCreated();
+		assertEquals(new Date(2018,5,10).getTime(), date.getTime());
```
junkerm: Could you please document that this is the default value fixed in the migration? You could also use a constant for this.

junkerm: to put it differently, I had a hard time figuring out why the assertEquals will be true. It took me while to see that this is the same value as used in the migrator. If you could make this a field and reference it from the migrator and the test it would be immediatly clear. As an alternative just add a comment.

munterkalmsteiner: Oh yes, of course. Fixed!

PR #155, opened 2018-05-24
Project aware search

junkerm:
Restricts search to the current project


bundles/specmate-dummy-data/src/com/specmate/dummydata/DummyDataService.java
```diff
@@ -53,28 +53,44 @@ public void setLogService(LogService logService) {
...
+					} catch (Exception e) {
+						logService.log(LogService.LOG_ERROR, e.getMessage());
+					}
+				}
+			} catch (SpecmateException e) {
+				logService.log(LogService.LOG_ERROR, "An error occured while inserting dummy data:", e);
 			}
-		}
+		}).start();
 	}
```
munterkalmsteiner: I think these timing related workarounds should be better documented:
- what exactly caused the workaround?
- what are the conditions under which the workaround can be removed?

In my opinion, this should be done for each Thread.sleep(x) instance in the complete code base as new developers (or we ourselves in a year) won't be able to trace all timing related issues to their root causes.

munterkalmsteiner (changes requested):
The search does not work for me. Also, the integration test fails with a 404 response.

munterkalmsteiner:
I've tried the search function, but I don't get any results... can you confirm that the search works for you (in the specmate GUI)?

junkerm:
@munterkalmsteiner in fact it does not work in the UI, probably there is still a timing problem involving the dummy data and the indexer, I look into it.

junkerm:
@munterkalmsteiner again I added the delay to the dummy data service. This seems to fix the issue. I added comments to clearify the rationale.

bundles/specmate-emfrest/src/com/specmate/emfrest/internal/rest/SpecmateResource.java
```diff
@@ -96,6 +108,31 @@ public final Object delete(@PathParam(SERVICE_KEY) String serviceName) {
 				true);
 	}
 
+	private boolean checkProject(EObject update, boolean recurse) {
```
munterkalmsteiner: Maybe it would be good to rename the method to something more descriptive, like isProjectModificationRequestAuthorized (or something more concise) and then add a comment how we check authorization (basically, by ensuring that the project from which the modification request origins corresponds to the to be changed project). Also, add a brief explanation why the recursion is necessary.

munterkalmsteiner (changes requested):
Search is still not working for me. When I type in the search field, the results field stays empty. Can you confirm that it works for you?

munterkalmsteiner (approved):

munterkalmsteiner:
Search worked again after deleting stale database files.

PR #160, opened 2018-05-27
Support for more fine-grained access rights control

munterkalmsteiner:


bundles/specmate-auth/src/com/specmate/auth/api/IAuthenticationService.java
```diff
@@ -1,10 +1,13 @@
...
 
 import com.specmate.common.SpecmateException;
+import com.specmate.usermodel.AccessRights;
 
 public interface IAuthenticationService {
 	public String authenticate(String username, String password, String projectname) throws SpecmateException;
 	public String authenticate(String username, String password) throws SpecmateException;
 	public void deauthenticate(String token) throws SpecmateException;
 	public void validateToken(String token, String path, boolean refresh) throws SpecmateException;
+	public AccessRights getALMAccessRights(String token) throws SpecmateException;
```
junkerm: I would suggest we be more generic here: Instead of PPM and ALM I would prefer source system and target system, as in the future it could be something else (jira, trello, etc.)

bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/AccessRights.java
```diff
@@ -0,0 +1,60 @@
...
+		String token = queryParams.getFirst(TOKEN_PARAM); 
+		
+		List<String> serviceParams = queryParams.get(SERVICE_PARAM);
+		if (serviceParams.isEmpty()) {
+			return Response.status(Response.Status.BAD_REQUEST).build();
+		}
+		String service = serviceParams.get(0);
+		
+		try {
+			if (service.equals("ALM")) {
```
junkerm: The same as above, I would rather term this "source" and "target"

PR #161, opened 2018-06-03
Project selection

munterkalmsteiner:


bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/ProjectNames.java
```diff
@@ -0,0 +1,36 @@
...
+	public static final String SERVICE_NAME = "projectnames";
+	private IProjectService projectService;
+	
+	@Override
+	public String getServiceName() {
+		return SERVICE_NAME;
+	}
+	
+	@Override
+	public boolean canGet(Object obj) {
```
junkerm: you could restrict this to the root "/" by checking if obj is of type Resource

bundles/specmate-integration-test/src/com/specmate/test/integration/support/DummyProjectService.java
```diff
@@ -21,6 +23,13 @@ public IProject getProject(String projectName) {
 		return projects.get(projectName);
 	}
 	
+	@Override
+	public List<String> getProjectNames() {
+		List<String> names = new ArrayList<>();
```
junkerm: instead of creating a new list every time, you could just use Set as return type and return directly the key set (may be a readonly view --> see Collections.asUnmodifiableSet)

bundles/specmate-emfrest/src/com/specmate/emfrest/authentication/ProjectNames.java
munterkalmsteiner: Ok

bundles/specmate-integration-test/src/com/specmate/test/integration/support/DummyProjectService.java
munterkalmsteiner: Yes, I changed that in the ProjectService API. However, the REST service still creates and returns a list since sets do not have a default marshal implementation in javax.

PR #164, opened 2018-06-08
Authentication in file connector

junkerm:
- Allows projects without exporter
- Allows to configure username and password with a file connector


bundles/specmate-file-connector/src/com/specmate/connectors/fileconnector/internal/FileConnector.java
```diff
@@ -161,6 +171,11 @@ public void setLogService(LogService logService) {
 
 	@Override
 	public boolean authenticate(String username, String password) {
-		return true;
+		if(user==null) {
+			return true;
```
munterkalmsteiner: Shouldn't the default case, i.e. when not properly configured, be to return false?

munterkalmsteiner (approved):
Looks good, minor comment on default behavior when user/password not configured.

PR #165, opened 2018-06-08
Export authorization check

munterkalmsteiner:
This implements authorization checks for export in the back-end.


munterkalmsteiner:
Correct, that will be a separate request in order to keep the PR's at a manageable size.

PR #167, opened 2018-06-11
Removed boldface styling from data-item-title

Daedo:
Minor change to conform to the software ergonomy standards.


junkerm:
What should change in the UI here? So that I can test it.

Daedo:
The change was intendet to make the UI align more with our usability styleguide.

junkerm:
In the Links&Actions panel the requirements text is now hard to distinguisch from the label "Anforderungsbeschreibung". Maybe add a linebreak?

PR #168, opened 2018-06-12
Gui export check

munterkalmsteiner:
We use the session information to decide whether the export on the client side should be allowed or not.


munterkalmsteiner:
I'll rename all ALM named files, strings, etc. in a new PR as there are several places where this needs to be fixed.

PR #169, opened 2018-06-13
Ppm proxy interface update

junkerm:
This pr updates the hp connector to work with the updated interface of the HP proxy


munterkalmsteiner (changes requested):
I was not able to add the comment to the line in the code where it belongs (Github gives me the following error message: "Start commit oid is not part of the pull request"), so I write it here.

In file ConnectorService, you change the comparison in the while loop form <= to <.
However, with this change the loop will not be executed if there is only 1 requirement in the list.

It's in commit 95261342afda33fc4a64f45117cc4d3b45551357

PR #175, opened 2018-06-17
History show actual user

munterkalmsteiner:
Storing now the user name in the transaction comment. Creating, storing, and displaying of change comments will come in a later PR.


bundles/specmate-persistency-api/src/com/specmate/persistency/ITransaction.java
```diff
@@ -1,43 +1,45 @@
...
 	boolean isDirty();
 
 	/**
 	 * Perform a change and commit
-	 * 
+	 *
 	 * @throws SpecmateValidationException
 	 */
-	<T> T doAndCommit(IChange<T> change) throws SpecmateException, SpecmateValidationException;
+	<T> RestResult<T> doAndCommit(IChange<T> change) throws SpecmateException, SpecmateValidationException;
```
junkerm: The transaction should be indedependent of the used interface. I would not use RestResult as a return type

munterkalmsteiner: Ok, reverted at the cost of manual type casting.

PR #180, opened 2018-06-22
display inconsistent test cases

junkerm:


junkerm:
Removing duplicate inconsistent tests. The issue with the "wrong" inconsistent tests will be handled in a separate PR as it is a larger issue. E.g. we need to display intermediate node values.

Daedo:
Okay, then we maybe should open an issue for that.

PR #189, opened 2018-06-29
Added missing i18n words for deletion of elements.

Daedo:


Daedo:
Just added one entry I missed.

PR #209, opened 2018-07-06
update to new hp proxy interface

junkerm:


bundles/specmate-hp-connector/src/com/specmate/connectors/hpconnector/internal/services/HPExportService.java
```diff
@@ -17,6 +17,7 @@
 public class HPExportService implements IExportService {
 
 	private HPProxyConnection hpConnection;
+	private String projectName;
```
munterkalmsteiner: Looking at the HPProxyConnection, it seems weird that the export check doesn't require a project name while the read check does. Having here an unused variable makes this even more suspicious. Is this implemented as intended? If yes, please remove the unused variable.

junkerm: The option to supply the project is not supported by the HP Proxy yet (but will in future). I removed the unused variable.

PR #210, opened 2018-07-06
add test for delete and search

junkerm:


bundles/specmate-integration-test/src/com/specmate/test/integration/SearchTest.java
```diff
@@ -186,6 +186,10 @@ public void testSearch() throws InterruptedException {
 		foundObjects = performSearch(projectFolderId, "bla -");
 		Assert.assertEquals(0, foundObjects.length());
 
+		deleteObject(projectFolderId, requirementId);
```
munterkalmsteiner: I'd add an assert that the object actually existed and was found by the search before the delete operation.
foundObjects = performSearch(projectFolderId, "BLUP");
Assert.assertEquals(1, foundObjects.length());

junkerm: This test is present, in line 131

PR #211, opened 2018-07-06
fix: id not shown in tracing view

junkerm:
fix: search in tracing view needs full words


web/src/app/modules/views/side/modules/tracing-links/components/tracing-links.component.html
```diff
@@ -5,11 +5,11 @@ <h5 class="card-header">
     </h5>
   <div class="card-body" [ngbCollapse]="isCollapsed" >
     <ng-template #rt let-r="result">
-      <i class="fa fa-plus-circle" aria-hidden="true"></i>&nbsp;<span *ngIf="extId">{{extId}}:&nbsp;</span>{{r.name}}
+      <i class="fa fa-plus-circle" aria-hidden="true"></i>&nbsp;<span *ngIf="r.extId">{{r.extId}}:&nbsp;</span>{{r.name}}
```
sebeder: r.extId leads to a warning in the template, since extId is not necessarily a field of r. Thus, you should use r['extId'].

PR #237, opened 2018-07-24
Using browser focus for keyboard navigation

sebeder:


sebeder:
This MR is a little more than only keyboard navigation; I also changed the welcome page and fixed the navigation-on-save-bug. Note that keyboard-navigation in language chooser will now only work with ```Tab``` and ```Enter```.

web/src/app/modules/navigation/modules/project-explorer/components/element-tree.component.ts
```diff
@@ -174,12 +181,59 @@ export class ElementTree implements OnInit {
...
 
-    public setSelection(): void {
-        this.treeNav.setSelection(this.baseUrl);
+    private move(off: number, elem: Element): void {
+        const canfocus = $(':focusable');
+        const currentIndex = canfocus.index(elem);
+        let index = currentIndex + off;
+        if (index >= canfocus.length) { index = 0; }
+        const targetElem = canfocus.eq(index);
+        console.log(elem);
```
Daedo: We can probably remove the logs.

sebeder:
I cannot reproduce the errors you found, @Daedo . You should always be able to navigate with up and down, and toggle with left/right/space/enter. What exactly dies not work?

Daedo:
Pressing space when an folder is selected causes it to toggle open and close it immidiately. Pressing arrow up while a folder is selected causes the focus to jump back to the safe button.

sebeder:
Interesting! Which browser are you using? I cannot reproduce this in chrome at least.

Daedo:
I'm using Firefox 61.0.1

sebeder:
OK, we do not support Firefox; Just Chrome and IE. Can you reproduce the error on Chrome?

PR #240, opened 2018-07-24
Graphical editor refactoring

Daedo:
### Editor Overhaul:

**Optimization**
Only Draw Visible Objects.
Optimize CEG Arc drawing.

**Multiselect**
Allow users to select and move multiple objects using click + drag rectangles.
Use SHIFT + Click to toggle selected elements.

**Keyboardshotcuts**
Use CMD + X, CMD + C, CMD + V to cut, copy, and paste selections (even between models as long as they have common type).
Backspace to delete selected elements.

**General Usabillity**
Hold SHIFT while using any tool to keep the tool active for multiple uses.
Use DOUBLECLICK to toggle CEG Connections negated/unnegated and to toggle CEGNodes AND/OR.


Daedo:
Doubleclick feature was removed by request.

web/src/app/modules/views/main/editors/modules/graphical-editor/components/process/process-end-graphical-node.component.ts
```diff
@@ -6,6 +6,8 @@ import { ProcessStep } from '../../../../../../../../model/ProcessStep';
 import { SpecmateDataService } from '../../../../../../../data/modules/data-service/services/specmate-data.service';
 import { SelectedElementService } from '../../../../../../side/modules/selected-element/services/selected-element.service';
 import { ValidationService } from '../../../../../../../forms/modules/validation/services/validation.service';
+import { MultiselectionService } from '../../../tool-pallette/services/multiselection.service';
+import { DomSanitizer } from '../../../../../../../../../../node_modules/@angular/platform-browser';
```
sebeder: This import is incorrect; "node_modules" should not be there.

web/src/app/modules/views/main/editors/modules/graphical-editor/elements/draggable-element-base.ts
```diff
@@ -51,20 +54,37 @@ export abstract class DraggableElementBase<T extends ISpecmatePositionableModelO
         this._rawY = y;
     }
 
+    private userIsDraggingElsewhere = true;
+
+    public get style() {
```
sebeder: I think it's better to solve this with different css classes. This way, we could also remove the reference to the sanitizer.

Additionally, this somehow leads to an error in IE, because the fill value is not interpreted correctly (see general review comments). Additionally, in IE, I had to set the fill value to rgba(0, 0, 0, 0.01) so that the drag events are emitted to the drag handle. With complete transparency (0 alpha), the click did not go to the drag handle.

web/src/app/modules/views/main/editors/modules/tool-pallette/util/graphTransformer.ts
```diff
@@ -0,0 +1,134 @@
...
+        // Old URL -> New Node map
+
+        let chain: Promise<any> = Promise.resolve();
+
+        for (const template of templates) {
+            if (this.elementProvider.isNode(template)) {
+                let temp = <IModelNode> template;
+                let newCoord = { x: temp.x, y: temp.y + 100};
+                let factory = GraphElementFactorySelector.getNodeFactory(template, newCoord, this.dataService);
+                chain = chain.then(() => factory.create(this.parent, false))
```
sebeder: In this case, I think it would be beneficial not to chain the promises manually, but use async/await to make the code more readable.

web/src/app/modules/views/main/editors/modules/tool-pallette/tools/drag-and-drop-tool-interface.ts
```diff
@@ -0,0 +1,7 @@
+import { ToolBase } from './tool-base';
+
+export interface DragAndDropToolInterface extends ToolBase {
```
sebeder: We usually name interfaces IBlaBlabla. Please adhere to this naming pattern.

web/src/app/modules/views/main/editors/modules/tool-pallette/tools/keyboard-tool-interface.ts
```diff
@@ -0,0 +1,5 @@
+import { ToolBase } from './tool-base';
+
+export interface KeyboardToolInterface extends ToolBase {
```
sebeder: Please name this interface IKeyboardTool.

PR #257, opened 2018-08-23
Improve test generation from CEG

junkerm:
Given two nodes in the form myvar=A, myvar=B, no test is generated with
myvar=A and myvar=B


munterkalmsteiner (approved):
Looks good to me (even though I cannot say much on the SAT solver code).

PR #259, opened 2018-08-31
Use own scheduler

sebeder:


bundles/specmate-config/config/specmate-config.properties
```diff
@@ -35,10 +35,14 @@ h2.jdbcConnection = jdbc:h2:./database/specmate
...
-## See: http://www.sauronsoftware.it/projects/cron4j/manual.php
+## generic value (will trigger every hour): hour
+## example: day 13 14 5 will trigger every day at 13:14:05
+## example: hour 14 5 will trigger every hour at xx:14:05
+## example: minute 5 will trigger every minute at xx:xx:05
+## missing numbers are replaced by 0s.
+## example: day 13 will trigger every day at 13:00:00
 # connectorPollSchedule = disabled
-connectorPollSchedule = * * * * *
+connectorPollSchedule = day
```
munterkalmsteiner: I think here day needs to be followed by at least one number to define the hour, otherwise an exception is thrown.

PR #265, opened 2018-09-20
Library information

munterkalmsteiner:


junkerm (changes requested):
Do we need a migrator, in order to reflect the new field in the session?
Otherwise, approved! :-)

PR #266, opened 2018-09-20
Validation service

munterkalmsteiner:


bundles/specmate-integration-test/src/com/specmate/test/integration/AttributeValidationServiceTest.java
```diff
@@ -0,0 +1,188 @@
...
+import org.osgi.util.tracker.ServiceTracker;
+
+import com.mifmif.common.regex.Generex;
+import com.specmate.common.SpecmateException;
+import com.specmate.common.SpecmateValidationException;
+import com.specmate.model.base.BaseFactory;
+import com.specmate.model.base.Folder;
+import com.specmate.model.support.api.IAttributeValidationService;
+
+public class AttributeValidationServiceTest {
```
junkerm: We could implement this test as a regular Junit test, which would run faster as an integration test as we do not need to start the whole framework

munterkalmsteiner:
@junkerm finished with the validation service implementation as we have discussed it earlier (closely tied to transactions).

bundles/specmate-connectors/src/com/specmate/connectors/config/ProjectConfigService.java
```diff
@@ -228,6 +223,10 @@ public Object doChange() throws SpecmateException {
 			String projectLibraryKey = PROJECT_PREFIX + projectName + KEY_PROJECT_LIBRARY;
 			String[] libraryFolders = configService.getConfigurationPropertyArray(projectLibraryKey);
 			if (libraryFolders != null) {
+				Hashtable<String, Object> projectLibraryConfig = new Hashtable<String, Object>();
```
junkerm: This is probably incorrect. I guess what should be done here is to retrieve an existing configuration and update it. However, as the project services are created as configuartion factory, a new configuration is created with the library folders as only information. This will spawn a new project service without project name, whereas the original project service will not be updates with the library folders.

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/BaseValidator.java
```diff
@@ -0,0 +1,23 @@
+package com.specmate.persistency.validation;
+
+import org.eclipse.emf.ecore.EObject;
+
+import com.specmate.model.base.INamed;
+import com.specmate.persistency.IChangeListener;
+
+public abstract class BaseValidator implements IChangeListener {
```
junkerm: rename to ValidatorBase (that is the convention that we used in the past)

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/LibraryFolderValidator.java
```diff
@@ -0,0 +1,55 @@
...
+import com.specmate.persistency.event.EChangeKind;
+
+public class LibraryFolderValidator implements IChangeListener {
+
+	// NOT NEEDED?
+
+	@Override
+	public void changedObject(EObject object, EStructuralFeature feature, EChangeKind changeKind, Object oldValue,
+			Object newValue, String objectClassName) throws SpecmateValidationException {
+
```
junkerm: shouldn't valdidateNotTopLevelLibraryFolder be called? If not, this class is in fact not needed

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/NameValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.specmate.common.SpecmateValidationException;
+import com.specmate.model.base.BasePackage;
+import com.specmate.model.base.INamed;
+import com.specmate.persistency.IChangeListener;
+import com.specmate.persistency.event.EChangeKind;
+
+public class NameValidator extends BaseValidator implements IChangeListener {
+	/** Pattern that describes invalid object names */
+	private static Pattern inValidNameChars = Pattern.compile("[,;|]");
```
junkerm: Why are these symbols not valid for names?

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/LibraryFolderValidator.java
munterkalmsteiner: I forgot to delete this class. The validation is already covered by TopLevelValidator since it is not necessary to distinguish between project top-level or library top-level folder; both have the same restrictions, are only applied in different contexts (which are controllable by when the validator is added to the transaction).

Note that TopLevelValidator is currently not used as it completely breaks the integration tests, which were not written with that constraint in mind. Adapting the tests is a major undertaking. I added a task to Trello, but I would not wait to merge this code because of that.

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/NameValidator.java
munterkalmsteiner: We use these in transaction commits for field and data separators. I added a comment for clarification.

munterkalmsteiner:
@junkerm fixed all comments, as far as possible in this PR. The tests are implemented as integration tests in CDOPersistencyValidationTest

bundles/specmate-connectors/src/com/specmate/connectors/config/ProjectConfigService.java
junkerm: no that is not it. The problem is that PROJECT_PID is not the PID of the service isntances but of the factory. The pids of the instances are automatically created by osgi and look like [PROJECT_PID]-[SOME-NUMBER]. Probably the right way would be to pick the service instance with a filter using the factory pid (PROJECT_PID) and the project name. 
Have a look here: https://stackoverflow.com/questions/30751407/how-can-single-instance-of-an-osgi-factory-configuration-be-read-from-java-in-cq

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/BaseValidator.java
junkerm: perfect

bundles/specmate-persistency-cdo/src/com/specmate/persistency/validation/LibraryFolderValidator.java
junkerm: I agree

bundles/specmate-connectors/src/com/specmate/connectors/config/ProjectConfigService.java
munterkalmsteiner: Ok, I see. Can you check 3192ae4 if that solves the bug?

munterkalmsteiner:
@junkerm please check out 966ce8abd07a21061a067152a4cfb6e8f168dfb1.

junkerm:
> 
> 
> @junkerm please check out [966ce8a](https://github.com/junkerm/specmate/commit/966ce8abd07a21061a067152a4cfb6e8f168dfb1).

junkerm:
Looks good! Go ahead and merge! :-)

PR #271, opened 2018-09-27
migration for string lists

junkerm:


bundles/specmate-dbprovider-h2/src/specmate/dbprovider/h2/AttributeToSQLMapper.java
```diff
@@ -97,14 +97,23 @@ public void migrateNewDateAttribute(String objectName, String attributeName, Dat
...
+	public void migrateNewStringReference(String objectName, String attributeName) throws SpecmateException {
+		migrateNewReference(objectName, attributeName, "VARCHAR(32672)");
+	}
+	
+	private void migrateNewReference(String objectName, String attributeName, String type) throws SpecmateException{
 		String failmsg = "Migration: Could not add column " + attributeName + " to table " + objectName + ".";
 		String tableNameList = objectName + "_" + attributeName + "_LIST";
 		List<String> queries = new ArrayList<>();
-		queries.add("ALTER TABLE " + objectName + " ADD COLUMN " + attributeName + " INTEGER");
+		queries.add("ALTER TABLE " + objectName + " ADD COLUMN " + attributeName + " " + type);
```
munterkalmsteiner: I'm wondering whether this change is correct as it modifies the behavior w.r.t. the previous implementation, even though it is intended to make the code more flexible (i.e. create generic references). 
Here, for the "ALTER TABLE..." query, the new column type was INTEGER when a new reference is created. Now it is BIGINT. I think it should stay INTEGER, i.e. not be parameterized, analogous to the oracle implementation. 

If this is indeed a bug, could you also check why the test cases do not capture it?

bundles/specmate-migration-test/build.properties
```diff
@@ -0,0 +1,9 @@
+#
+
+bin.includes = specmate-migration-test.jar,\
+               model/,\
+               plugin.xml,\
+               plugin.properties
+jars.compile.order = specmate-migration-test.jar
+source.specmate-migration-test.jar = src/
+output.specmate-migration-test.jar = bin/
```
munterkalmsteiner: Generated file, please remove

bundles/specmate-migration-test/plugin.properties
```diff
@@ -0,0 +1,4 @@
+#
+
+pluginName = Attributeadded Model
+providerName = www.example.org
```
munterkalmsteiner: Generated file, please remove

bundles/specmate-migration-test/plugin.xml
```diff
@@ -0,0 +1,40 @@
...
+
+   <extension point="org.eclipse.emf.ecore.generated_package">
+      <!-- @generated attributeadded -->
+      <package
+            uri="http://specmate.com/1/testmodel/artefact"
+            class="com.specmate.migration.test.attributeadded.testmodel.artefact.ArtefactPackage"
+            genModel="models/attributeadded.genmodel"/>
+   </extension>
+
+</plugin>
```
munterkalmsteiner: Generated file, please remove

munterkalmsteiner (changes requested):
Overall, looks good. One comment though that might need addressing.

PR #273, opened 2018-10-02
Fix "Unexpected unknown type selector "element-tree""

fswt:
## What was done
It is reported as a bug, because it is a unknown html type selector; instead it is generated by angular. The profile SonarQube uses to check for bugs in CSS files seems to be working with a dictionary of type selectors only, rather than check the selectors in the code.
We think it should be fixed, because as it appears it was a bug in Angular 2 and it wouldn't work before, although now it works with the current angular version, we are not sure how the behaviour is going to be on different browser or older versions of the browsers.
The fix is straight forward, instead of putting the style in the CSS file on the "element-tree" selector, we changed it to a class ".element-tree" and added it to the "element-tree" in the HTML file.


munterkalmsteiner:
Thanks for the pull request! Two things before I can merge:
1) There are two commit. The first is the merge of an old pull request (#8) into this branch. Don't know, why. I don't think you can remove this merge, so I think you need to make a new pull request in order to get rid of it.
2) Please make sure that your pull request does not cause merge conflicts with develop.

fswt:
Thanks for the feedback! We found some answers and solutions:
Concerning 1.: The reason of the merge of pull request #8 is that we wanted to update our repository (fswt/specmate) before opening a pull request to junkerm/specmate. Therefore we created a pull request (#8) in our repository, to merge the new commits from junkerm/specmate. So #8 actually refers to the pull request at fswt/specmate, although here it is referred to #8 of this repository.
Concerning 2.: We merged the changes into the pull request branch.

PR #278, opened 2018-10-02
Refactoring for same implementation in different methods

HeXiangtian:
Extract the same implementation part from different methods, create a new method to achieve this implementation, the original methods calling this new method to achieve original implementation.
Group 1


Wen1234:
The functions that we change are select and click functions in graphical-editor. We have manually test the functions and it can work as before.

HeXiangtian:
Change setActivateDefaultTool function from public to private

munterkalmsteiner:
I know what needs to be changed. Please fix and update the pull request.

HeXiangtian:
I'm sorry I forgot to push the request before, now you can see the change.

munterkalmsteiner:
I am sorry, this pull request is too messy... I don't see at all what you have changed and what is merged from somewhere else... Please open a new pull request.

PR #279, opened 2018-10-02
Refactor this function so that its implementation doesn't duplicate t…

Wen1234:
…he one on line 83

Because InputParameters and OutputParameters have the same implementation, so I create getInOutParameters to return outParameters to call. The functions that we change are inputParameters and outputParameters functions in Test-case-row. We have manually test the functions and it can work as before.
Group 1


Wen1234:
I remove the duplicated code in method inputParameters

PR #282, opened 2018-10-02
Fix group 8 assigned issues

roostaamir:
This patch will resolve the issues that were assigned to group 8. The issues that were assigned to group 8 are the following:

- 3-2 Remove this use of the output from "updateElementVirtual"; "updateElementVirtual" doesn't return anything. (L)
- 3-3 Remove this use of the output from "deleteElementVirtual"; "deleteElementVirtual" doesn't return anything (L)

### DIFF
--- a/web/src/app/modules/data/modules/data-service/services/specmate-data.service.ts
+++ b/web/src/app/modules/data/modules/data-service/services/specmate-data.service.ts
@@ -225,12 +225,12 @@ export class SpecmateDataService {
         return this.cache.readElement(url);
     }
 
-    private updateElementVirtual(element: IContainer, compoundId: string): void {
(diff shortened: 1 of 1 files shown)


roostaamir-knowit:
Actually, we believe that this case was a false positive that Sonar was giving us. At least at its current context. However, making the definition more generic, the Promise object now is resolvable for a broader context, hence Sonar will not give an error for it. If we had a build system that prevented builds on errors, this would prevent us from getting correct builds.

roostaamir-knowit:
Yes we do believe that it is a false positive

munterkalmsteiner:
Ok, thanks! Then I close this PR.

PR #283, opened 2018-10-02
Changed fields function cherry pick

fswt:
## What was done
Reduced the length and cognitive complexity of the changedFields function. Also added some comments, to improve understandability. 
In addition another bug was found: The code would have crashed, if the second object (see changedFields) did not contain something that the first object did. Fixed that issue.

## Testing
We created some tests to check our changes, which can be found [here](https://github.com/fswt/specmate/blob/developGroup7/web/src/app/util/objects.spec.ts).
The results of the tests can be examined [here](https://travis-ci.org/fswt/specmate).
For the testing we used [Mocha](https://mochajs.org/) and [Chai](https://www.chaijs.com/).


MindTraper:
We changed the tests (old link still works) and will send a new pull request. The reason that 2  "___nsuri" appeared , is that inside our function "pushTheNotMatchingFields" we had an if statement "if(!o1[field])" which will return false if the string is empty. Since we have 2 iterations, one for each object , that is why it appeared twice. Now we add value in the "___nsuri" field and the test runs correctly and just id is returned.

fswt:
We will create a new pull request, because here we also have this strange pull request merge in the commit history.

fswt:
Correction of the comment above: If not required, we will not open a new pull request, as the merge of pull request #8 is wanted as explained in #273 .

munterkalmsteiner:
I'm sorry, but the Travis output does not show that the tests have been run... could you check that please?

fswt:
Should be fixed. It shows it now again.

munterkalmsteiner:
The first test is disabled so I don't know if it actually works as expected...
// expect(Objects.changedFields(o1, o2)).to.be.undefined;

Sorry for being a PITA...

fswt:
We added the line again. 
We removed it before due to some issues with "build-dev", but as the tests will not be merged into your repository this just concerned our fork.

PR #289, opened 2018-10-04
Library folder - delete and edit properties

munterkalmsteiner:
Library contents (sub-folders, models, etc.) are shown in the middle pane and properties (name, description) can be edited on the right pane. History is also shown in the right pane. 
The missing information in the history (who deleted/created what) I would address not here but in a second PR.

There is still a subtle bug: after selecting a subfolder or model, coming from a top-level library folder, the properties of the navigated to object are not visible (no element selected).


munterkalmsteiner:
Fixed bug and made requested changes. Merging.

PR #299, opened 2018-10-16
Focus on node's main property when selected

roostaamir:
- Feature number 6 in the list (group 8)
- When an element is selected in in the graphical editor, it main
  property will be focused in the properties window so that the user
  can start giving values to it immedietly


sebeder:
Hi everyone!

From a functional point of view, the behavior is fine.

However, the implementation is a little fragile: You are using the id to find the actual control you want to focus. There might be a more robust way using the ```@ViewChild```-annotation in the property editor and in the generic input form to access the acual DOM elements.

But as I just said, the second solution is hard to implement (if it works at all). That's why I'd accept the proposed solution.

munterkalmsteiner:
Thanks @sebeder! 
@roostaamir, do you think you could implement the approach proposed by @sebeder?

munterkalmsteiner:
Merging based on the [review](https://github.com/junkerm/specmate/pull/299#issuecomment-433400312) by @sebeder.

PR #301, opened 2018-10-18
Removed the duplication "confirmDelete" function on line 18

sriram7386:
The functionalities "confirmdelete" are being called by another function "openOkcancel" which results in duplication. The code has been refactored by changing the "openOkcancel " recurrences to "confirmDelete". The functionalities of "openOcancel" are not being used anywhere but only in this code segment. Duplication has been removed as code clone incur the maintainability costs and effort.

### DIFF
--- a/web/src/app/modules/notification/modules/modals/services/confirmation-modal.service.ts
+++ b/web/src/app/modules/notification/modules/modals/services/confirmation-modal.service.ts
@@ -27,15 +27,9 @@ export class ConfirmationModal {
         return modalRef.result;
     }
 
-    public openOkCancel(title: string, message: string): Promise<any> {
-        const modalRef = this.modalService.open(TypedModalContent);
-        modalRef.componentInstance.options = Dialogtype.okCancelDialog(title, message);
-        return modalRef.result;
-    }
-
     public confirmSave(message?: string): Promise<void> {
         if (this.dataService.hasCommits) {
-            return this.openOkCancel('ConfirmationRequired', message || this.translate.instant('confirmSave'));
+            return this.confirmDelete('ConfirmationRequired', message || this.translate.instant('confirmSave'));
         }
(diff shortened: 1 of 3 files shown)


web/src/app/modules/notification/modules/modals/services/confirmation-modal.service.ts
```diff
@@ -16,18 +16,18 @@ export class ConfirmationModal {
...
         const modalRef = this.modalService.open(TypedModalContent);
         modalRef.componentInstance.options = Dialogtype.okDialog(title, message);
         return modalRef.result;
     }
-
-    public openOkCancel(title: string, message: string): Promise<any> {
+  public openOkCancel(title: string, message: string): Promise<any> {
+        return this.change(title,message);
+    }
+    public change(title: string, message: string): Promise<any> {
```
munterkalmsteiner: This method name needs to be more informative. Change to openOkCancel.
Also, there is no need for this method to be public.

web/src/app/modules/views/main/authentication/modules/login/components/login.component.css
```diff
@@ -1,7 +1,12 @@
 .not-allowed {
     cursor: not-allowed;
 }
+.bgimg
```
munterkalmsteiner: Please create 1 pull request per feature (revert this change and open a new pull request).

munterkalmsteiner:
Please see my previous comments. I cannot merge before they are not addressed.

web/src/app/modules/views/main/authentication/modules/login/components/login.component.css
manpreethb: We have placed a new pull request titled "Feature-7 (Group-3)Background image for login page ".
Thank you.

PR #303, opened 2018-10-18
Fix unnecessary boolean

pave17:
### What we done?

deleted three unnecessary boolean , it was returning value && true, so since the return value always depended on the other value so it is unnecessary and make confusion.

group 6


pave17:
Am i need to remove branch? or I need to send the new pull request with a creating a new branch?

munterkalmsteiner:
You need to create two branches and two separate pull requests. For each pull request, describe also briefly in the description what is actually fixed. You did that correctly for the unnecessary boolean but I don't know what the font-family change should fix.

PR #304, opened 2018-10-18
Removed undefined

pave17:
What we have done?
  removed undefined from the code.


munterkalmsteiner:
This pull request addresses different issues.
Create new pull requests, each addressing one and only one issue.

PR #305, opened 2018-10-18
Refactor the get to refer Requirement and model

manojkumarpallapu:


web/src/app/modules/views/side/modules/links-actions/components/links-actions.component.ts
```diff
@@ -16,8 +16,8 @@ export class LinksActions {
 
     public isCollapsed = false;
 
-    public _requirement: Requirement;
-    public _model: IContainer;
+   // public _requirement: Requirement;
+   // public _model: IContainer;
     public _contents: IContainer[];
```
munterkalmsteiner: Better to delete than to comment.
Also, it seems that _contents and _testSpecifications are not used anywhere. Why not remove them too?

web/src/app/modules/views/side/modules/links-actions/services/additional-information.service.ts
```diff
@@ -17,8 +17,8 @@ import { UserToken } from '../../../../main/authentication/base/user-token';
 @Injectable()
 export class AdditionalInformationService {
 
-    private _requirement: Requirement;
-    private _model: IContainer;
+    public _requirement: Requirement;
+    public _model: IContainer;
```
munterkalmsteiner: Why? I don't understand this change... wouldn't it be better to delete _requirement, _model, and _contents as they are not used?

PR #308, opened 2018-10-23
Making text multiline

sebeder:
The reason is that we need more space for the text displayed in graphical nodes. So, I made the text in the nodes using multiple lines.


Daedo:
It seems to work fine however in many cases I've tried it dosn't look very nice.
The text doesn't get centered vertically and the variable name is too close to the top. 
Having a short variable name and a long condition makes the condition text get close to the bottom of the frame while leaving an empty line above. This also means that if the condition text exceeds the two lines it gets truncated dispite there being enough space to write it out.
It might be a good idea to have the position of the text be more flexible in order to avoid these issues and make full use of the available space.

PR #310, opened 2018-10-26
Feature 13, disabled the button which are disabled, when pressing the…

ruanlx22:
It is about feature 13, Disabled button not reachable by pressing tab.

**Test case**
**Title:** Feature 13, Disabled button not reachable by pressing tab
**Description:**  After users logging into Specmate and do not do anything, the status of Save, Undo, Back, Forward button are disable, and it should be not reachable after implemating this feature when pressing tab-key.
**Test steps:**
1. Open specmate in browser.
2. Login into system.
3. Press tab-key and check wheather the dsiabled button could be reachable.

**Expect result:** the status of Save, Undo, Back, Forward button should be disabled and not reachable.

BR,
Group 2


munterkalmsteiner:
Thanks, the code is correct. However, the feature was already implemented by another group and I have already merged it, so I need to reject this pull request. Still, good work!

PR #315, opened 2018-10-26
Feature 12 (Mandatory) - Changed Tab Order buttons of Graphical Editor (Group 3)

manpreethb:
The tab order of the buttons in graphical editor was initially from right to left. We refactored the code so that it is changed from left to right. Only the order of the buttons is changed and the position of the buttons is still on the right top corner.


munterkalmsteiner:
Your solution looks correct, we have however received and merged an earlier pull request: #296. Anyway, thank you for your contribution!

PR #316, opened 2018-10-26
Feature 2; Export CSV; (Group 7 - Mandatory)

fswt:
# Additional Information
The required, additional information can be found [here](https://docs.google.com/document/d/1Zxr3vOVpPgepF71RrQd4OYnGfiHGA9qsf-6LbfXFs1g/edit?usp=sharing).


munterkalmsteiner:
@sebeder, this looks like a good implementation to me. Could you have a look at it too?

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+    private finalCsvString: string;
+
+    constructor(
+        private dataService: SpecmateDataService,
+        private validation: ValidationService,
+        private translate: TranslateService,
+        private auth: AuthenticationService) { }
+
+    // Export Function
+    public exportTestSpecification(): void {
```
sebeder: I'd make this method async to make it easier to handle all the asynchronous calls inside. Like this:

```suggestion
    public async exportTestSpecification(): Promise<void> {
```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+        if (!this.enabled) {
+          return;
+        }
+
+        this.dataService.readContents(this.testSpecification.url)
```
sebeder:
```suggestion
        const contents = await this.dataService.readContents(this.testSpecification.url);
        this.contents = contents;
```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+          .then((contents: IContainer[]) => {
+            this.contents = contents;
+            this.loadTestParameters();
```
sebeder: As far as I see, the called function is actually asnychronous (it returns a Promise). That's why we should ```await``` it like this (also the next function call):

```suggestion
            await this.loadTestParameters();
```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+            this.loadTestCases();
```
sebeder:
```suggestion
            await this.loadTestCases();
```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+            Promise.all(this.loadParameterAssignments()).then(() => {
```
sebeder:
```suggestion
            await Promise.all(this.loadParameterAssignments());
```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+              this.prepareExportString();
+              this.createDownloadFile();
+            });
```
sebeder:
```suggestion

```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+          });
```
sebeder:
```suggestion

```

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+                                this.parameterAssignments.push(element as ParameterAssignment);
+                          }
+                        }));
+            });
+            promiseArray.push(loadParameterAssignmentsTask);
+        }
+        return promiseArray;
+    }
+
+    private createDownloadFile(): void {
```
sebeder: Is there a library that can manage this for us? In my experience, doing all this manually is fragile at least. If there is a library, I am happy to put it into specmate. I think we will need it at some other points as well.

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
```diff
@@ -0,0 +1,181 @@
...
+        private auth: AuthenticationService) { }
+
+    // Export Function
+    public exportTestSpecification(): void {
+        if (!this.enabled) {
+          return;
+        }
+
+        this.dataService.readContents(this.testSpecification.url)
+          .then((contents: IContainer[]) => {
```
waleedabdeen: I guess this should be deleted as well to follow with the previous change.

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
waleedabdeen: @sebeder  can you give me more context. do you mean the way of preparing the string from the loaded arrays (pushRows and pushHeaders) or everything !? you want a library to take the current content array and give us the csv string ?

waleedabdeen:
@sebeder  i am taking care of all the await and async issues as you suggested , and had some other comments, if you could check it please.

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
munterkalmsteiner: @waleedabdeen, no, I think I know what @sebeder  means. Currently, this is the only place we need this feature, i.e. create a file on the client side such that the browser offers the user to open or save the file. However, we have some use cases in mind where we need that functionality in the future. It would therefore make sense to rely on a tested (on many platforms and browsers) solution that we don't have to maintain ourselves.
Maybe [FileSaver.js](https://github.com/eligrey/FileSaver.js) would work (it has almost as many upvotes as the the solution you chose from SO :P)

waleedabdeen: @munterkalmsteiner  thanks for the clarification, i will look into it and try to implement it.

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
waleedabdeen: agree, but just before i submit the changes will this be completed (readContents) before executing (loadTestParameters) ??? because loadTestParameters depends on (this.contents) variable.

sebeder: Yes, this is exactly what the ```await```-keyword is for. It's basically syntactic sugar to avoid all these ```promise.then(...)```-call chains. So what it does (simplified) is, that it wraps the code after the statement into a ```then(code below)```-call.

web/src/app/modules/actions/modules/export-testspecification-button/components/export-testspecification-button.component.ts
sebeder: Hi @munterkalmsteiner and @waleedabdeen I think we can need this for the test-case-generation (to allow people to actually download the text generated for test stubs). That's why I'd like to have it as a library.

waleedabdeen:
ok, could you please take a look at the latest changes, i have fixed the await , async and used the file-saver method (saveAs) to create the download file. other changes i should make ?

munterkalmsteiner:
@waleedabdeen, I will merge your changes. Thank you for your contribution!

PR #318, opened 2018-10-26
Feature-10(1) (Group-5):Change language Button size

sravanibejawada25:
For easy understanding of the functionality button size is increased.

PR #319, opened 2018-10-26
Feature 10 Mandatory (group5)

gannaanil:
for better understanding pf functionality size is increased


munterkalmsteiner:
See my comment in #318.

PR #314, opened 2018-10-26
Feature 7; Background image; (Group 7 - Optional)

fswt:
# Additional Information
The required, additional information can be found [here](https://docs.google.com/document/d/1hBLAKmNPDkRrmh_6mSA7PtDxYDgdJ3btB_AlEEdGqz8/edit?usp=sharing).

PR #320, opened 2018-10-26
Group 4 Feature 7 variant 1 (Mandatory)

anushamogili:
We have implemented using afterviewinit hook, which is invoked after angular has initialized components view. By using this method we have implemented background image for login page and it is not visible after logging into specmate app. (This is another method to give the background image.)
It is working properly in our system and background image is displayed. There is no Afterviewinit error in our system.


munterkalmsteiner:
The implementation in pull request #314 is more flexible and builds without errors. I chose therefore to close this pull request. Anyway, thank your for your contribution!

PR #321, opened 2018-10-26
Group 4 feature 7 variant 2 (mandatory) -: the user can select and co…

anushamogili:
…nfigure the background image. In addition, the user can upload several background images and on each reload of the login page, Specmate shows a random image

1.We have used input textbox with type file to upload images. An array is used to store the URL of uploaded images and background image of specmate is implemented.

2.  On each reload of the login page, Specmate shows a random image from the user uploaded images for this we have saved the array in session. When the page is reloaded the array images which are present in session is retrieved and displays randomly.

3.  Uploaded images are displayed on the right side of the page. We can also choose image from the uploaded images to set as background image. which is implemented in login.component.html(pull request 322)
4.Images should be uploaded again if tab is closed


munterkalmsteiner:
Sorry, build failure both on Travis and on my local machine. Closing.

PR #323, opened 2018-10-27
Removing undefined from the code Group6

pave17:
what we have done?
Removing undefined from the code.


munterkalmsteiner:
No. Each fix MUST be a separate branch and pull request. See my other comments.

PR #324, opened 2018-10-27
Fix unnecessary boolean

pave17:
What we have done?
Removing &&true from the code so that it works. 
Group 6


munterkalmsteiner:
Closing again... see my comment:
You need to create two branches and two separate pull requests. For each pull request, describe also briefly in the description what is actually fixed. You did that correctly for the unnecessary boolean but I don't know what the font-family change should fix.

Also, make sure your branch has pulled the latest version of develop from the specmate repository. This pull request has a merge conflict that would need to be resolved first.

PR #325, opened 2018-10-29
reusing ceg models

sebeder:


munterkalmsteiner (changes requested):
I have tested the behavior of copying a model. When pasting the model, the clipboard is not emptied, i.e. the model is still there and I can paste the model again and again (see attached screenshot)
screenshot_20181106_110754. 

I assume the expected behavior is that, once the model is pasted, the clipboard is empty. Note: this is different from what we have discussed during the Jour Fixe, where I suggested that the user can empty the clipboard without pasting.

sebeder:
Hi @munterkalmsteiner , I'm clearing the clipboard now. Can we merge?

munterkalmsteiner (approved):

PR #328, opened 2018-10-31
Removed the duplication "confirmDelete" function on line 18

SantoshKumarSK:
The functionalities "confirmdelete" are being called by another function "openOkcancel" which results in duplication. The code has been refactored by changing the "openOkcancel " recurrences to "confirmDelete". The functionalities of "openOcancel" are not being used anywhere but only in this code segment. Duplication has been removed as code clone incur the maintainability costs and effort.


munterkalmsteiner:
Hm, I think it would make more sense to merge the functionality of `openOkCancel` and `confirmDelete` into one public method `openOkCancel` and then change the *callers*. This change does superficially solve a problem, but the root cause is not addressed. Closing now, but you are welcome to submit another pull request with the solution I described above.

PR #329, opened 2018-11-05
Added missing fileds names to validation error message.

Daedo:


sebeder:
Hi @Daedo , please make the build work (if poassible), then merge.

PR #330, opened 2018-11-06
Feature 1:Print view for models

Wen1234:
We add "hide UI" button in card-header model. After clicking this button, all UI elements except model view are invisible.  And when clicking the button again, the UI will back to the original view. And we use the manual test to test this button. We try to hide the tool palette, but we fail.


Wen1234:
Test case Title: Feature 1 Print view for models
Description: As a user, I want the print view shows the model without any UI elements so that I can print model suitably.
Test Steps:
1. Click the “hide UI” button
2. Click the “print” button within the browser
3. Click the “restore UI” button
BR, Group 1

web/src/app/modules/views/controller/modules/view-controller/services/view-controller.service.ts
```diff
@@ -56,6 +57,18 @@ export class ViewControllerService {
         return this._isEditorMaximized;
     }
 
+    public printEditor(): void {
+        this._isEditorPrint = true;
+    }
+
+    public unprintEditor(): void {
+        this._isEditorPrint = false;
+    }
```
munterkalmsteiner: I think it would be better to combine the two methods above and pass a boolean parameter.

web/src/app/modules/views/main/editors/modules/print-button/components/print-button.component.ts
```diff
@@ -0,0 +1,29 @@
+import { Component } from '@angular/core';
+import { ViewControllerService } from '../../../../../controller/modules/view-controller/services/view-controller.service';
+
+/**
+ * This is the Specmate main component
+ */
```
munterkalmsteiner: This comment is not correct...

web/src/assets/i18n/de.json
```diff
@@ -139,6 +139,7 @@
 	"paste": "Einfügen",
 	"pasteAsName": "Mit Namen einfügen",
 	"payload": "Ladung",
+	"print": "drucken",
```
munterkalmsteiner: Should be "Drucken"

PR #332, opened 2018-11-08
Testspec skeleton generation

munterkalmsteiner:
Generating java and javascript test spec skeletons that can be downloaded and saved.


sebeder (changes requested):
Good work! It "almost" works, and there are only minor things that popped up:

- There is always the "AAA" in the generated test cases. This should be the name of the test specification, I think.
- The download-links are shown not only on test-specification, but also models and so on. This leads to errors. Whe should check whether we can display the links (I will do this in this branch).
- I'd style the links and the text more carefully (I will do this in this branch).

munterkalmsteiner:
@sebeder thanks for the review. I have fixed all issues (except the one about styling, you know best how to fix it), and some additional issues (invalid characters in class/method names).

sebeder (approved):

PR #337, opened 2018-11-20
Rotate colors in project explorer tabs

sebeder:
We had the request to change the colors in the project explorer tab, so that the active tab is blue and the inactive tab is grey.


junkerm (approved):
works!

PR #339, opened 2018-11-27
Making history context sensitive

sebeder:
See https://trello.com/c/6m8ZWMvD/276-kontext-abh%C3%A4ngiger-%C3%A4nderungsverlauf


munterkalmsteiner (approved):
Looks good! I have also removed again the distinction between simplified/extended history as we don't actually use it anymore and it may just cause confusion in the future.

PR #341, opened 2018-11-27
Showing testspec generation button only on models in requirements

sebeder:
See https://trello.com/c/Uv0UveAa/290-generierung-von-testspezifikationen-soll-nur-im-projekt-kontext-m%C3%B6glich-sein


junkerm (approved):
works as expected, code looks good

PR #345, opened 2018-11-30
object.ts

gannaanil:


gannaanil:
In the refactored code we have created conditions in such a way that if one the of the functions return false then it goes back to main function. If all three statement are true and return true then it start comparing the objects. Which increases the understandability of the code.

munterkalmsteiner:
This pull request combines several different fixes that do not correspond to the description in your comment. I can therefore not accept the pull request. You need to create ONE pull request per feature or refactoring.

PR #346, opened 2018-12-02
Informative error responses

munterkalmsteiner:
This patch straightens out how error messages are handled in the back-end and provides error codes that can be used in the front-end to generate translated error messages.

SpecmateExcpetion is now abstract and provides some base functionality. We have 3 concrete implementations:
1. SpecmateAuthorizationException: thrown when a request is not authorized to perform an operation. Caught in SpecmateResource where the corresponding ProblemDetail message is generated and send to the client.  
2. SpecmateValidationException: thrown when data sent from the client is invalid. Can be generic or caused by a specific validator (indicated by the error code). In the latter case, the ProblemDetail object will contain more information.
3. SpecmateInternalException: thrown when something in the back-end goes wrong. 

In all three cases, the front-end can decide, based on the received error code, what action should be taken and how the user should be informed. In general, error codes < 100 indicate a problem on the client (invalid data, authorization), and error codes >= 100 indicate a problem in the back-end.

Hence, the recommended strategy would be to provide translated error messages for error codes < 100 and provide the actual error code for error code >= 100, so the user can report that reference to the developers.


munterkalmsteiner:
@junkerm I implemented tests (and fixed some small issues resulting from the tests). I've implemented tests that trigger errors that can be triggered via the rest interface, i.e. only client errors. Internal errors are somewhat difficult to provoke from the client side, so I did not spend much time on investigating that. I think that is ok, since there is no inherent difference between client side and internal errors.

PR #347, opened 2018-12-03
One-off error

munterkalmsteiner:
The "to" parameter in copyOfRange is exclusive. The faulty implementation chopped off the last digit in the time parameter string.

I'm not sure if this causes the observed behavior at the client. If, for example, the scheduler was set to "day  03", the faulty implementation would have created a daily task at midnight, i.e. ignoring the "03". This *could* lead to the observed behavior if the scheduled task does not have the resources (e.g. network access) at midnight.


junkerm:
I think that was not the only error. It still didn't work for me, so I investigated. I found an error in the hourly and minute iterators. To be sure I added some unit tests for the iterators

PR #348, opened 2018-12-04
Ignoring changes, where an undefined field becomes an empty string

sebeder:
See: https://trello.com/c/5HT21CSi/313-unn%C3%B6tige-popups-bei-geister%C3%A4nderungen

Behavior:
- Reload Specmate
- Open a library folder
- Open another library folder
- Unsaved-changes modal is displayed (but there are no changes, actually)

Desired Behavior: No changes modal is displayed

This PR realizes the desired behavior.


munterkalmsteiner (approved):
Works as advertised!

PR #352, opened 2018-12-11
Specify if a folder is a library or not

munterkalmsteiner:
This fixes the issue described here:
https://trello.com/c/DT33S5Cr/253-bibliothek-top-level-ordner-%C3%BCber-projektkonfiguration-beschreiben


munterkalmsteiner:
@junkerm kannst du dir diesen PR anschauen? Danke!

bundles/specmate-migration/src/com/specmate/migration/internal/services/Migrator20181108.java
```diff
@@ -0,0 +1,74 @@
...
+			if (projectsNames != null) {
+				for (int i = 0; i < projectsNames.length; i++) {
+					String projectName = projectsNames[i];
+					String projectLibraryKey = IProjectConfigService.PROJECT_PREFIX + projectName
+							+ IProjectConfigService.KEY_PROJECT_LIBRARY;
+					String[] libraryFolders = configService.getConfigurationPropertyArray(projectLibraryKey);
+					if (libraryFolders != null) {
+						for (int j = 0; j < libraryFolders.length; j++) {
+							String libraryFolder = libraryFolders[j];
+							String sql = "UPDATE FOLDER set library = true where id = '" + libraryFolder + "'";
```
junkerm: Können wir sicher sein, dass die id eindeutig ist? Vermutlich nicht.
Können wir zumindest noch checken, dass der aktuelle Parent-Ordner ein (bzw. der richtige) Projektordner ist?

munterkalmsteiner: The migrator ensures now that only those folders are marked as library folders that:
1) are, at the time of migration, configured in a project as library folders
2) are indeed child folders of the configured project with library folders 

1) is a necessary constraint (i.e. we cannot mark as library folders all folders that are sub-folders of projects, in general, but only those that are configured as library folders). This needs to be considered when running the migration on a production system, e.g. by temporarily adding all removed library folders again to the configuration.

munterkalmsteiner:
@junkerm I've fixed the migrator and replicated your manual test. At least for me, when removing a library folder from the configuration, it does not show up in specmate. Can you confirm?

PR #356, opened 2018-12-17
Graph Validation

Daedo:
Added three new validators:
* Detects inner nodes that only have a single cause
* Detects nodes with contradictory causes
* Detects circular causes
Added Set & Array utility functions


sebeder (changes requested):
It's fine from a technical point of view - However, the code in the validators is quite long. Could you refactor this into smaller methods (and add comments) to increase readability?

web/src/app/validation/ceg/contradictory-condition-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+            toNode.parentCount--;
+            if (toNode.parentCount == 0) {
+                // We have explored all parents of toNode. It can now be explored.
+                // Test for contraditcions and add mark it for the openList.
+                for (const key in toNode.posEdgeSets) {
+                    if (toNode.posEdgeSets.hasOwnProperty(key) && toNode.negEdgeSets.hasOwnProperty(key)) {
+                        // We have a contradiction
+                        // Cause -PosPath-> Effect
+                        // Cause -NegPath-> !Effect
+                        let cause = this.topoNodes[key].node;
```
sebeder: Change `let` to `const`, since the values are not changed anymore.

web/src/app/validation/ceg/contradictory-condition-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+        this.posEdgeSets = {};
+        this.negEdgeSets = {};
+        this._setWasFilled = false;
+    }
+
+    private static cloneEdgeSet(edgeSet: {[parentURL: string]: CEGConnection[]}): {[parentURL: string]: CEGConnection[]} {
+        /**
+         * Helper function. Creates copy of an edge set. Also copys the arrays within the set.
+         */
+        let out = {};
```
sebeder: Also here, you can use `const` instead of `let` - please review your code regarding the use of `let` and `const`.

web/src/app/validation/ceg/contradictory-condition-validator.ts
```diff
@@ -0,0 +1,285 @@
...
+     * A CEG Node with all its outgoing edges.
+     * A counter for the number of unexplored parents (for the topological sorting)
+     * and the nodes' positive and and negative cause sets.
+     */
+    public node: CEGNode;
+    public edges: CEGConnection[];
+    public parentCount: number;
+    public posEdgeSets: {[parentURL: string]: CEGConnection[]};
+    public negEdgeSets: {[parentURL: string]: CEGConnection[]};
+    private _setWasFilled: boolean;
```
sebeder: We usually do not use underscores for private variables. You can remove it here as well.

sebeder (changes requested):
Please resolve the merge conflicts as well.

sebeder (changes requested):
I still think the validation code is complicated and hard to understand. Could you refactor it a little (break it up in single/smaller methods)? In addition to this, we cannot merge this right now, since we are in the middle of a testing phase and I don't want to jeopardize this.

Daedo:
I have done some further simplification. But I'm still not fully happy with the contradiction test. If you'd have any suggestions I'd be open for those.

sebeder (changes requested):
Just one more thing:
- Please add translations to error messages.

Then, feel free to merge.

sebeder (approved):

PR #363, opened 2018-12-28
fix bug where CEGNodes are found by search after a change to the node

junkerm:


junkerm:
@munterkalmsteiner Did you delete your database directory before testing? It might be that you have an old search index.

munterkalmsteiner:
@junkerm yes, that was the problem. After deleting the database, I cannot reproduce the issue. Please merge.

PR #364, opened 2018-12-28
fix bug that prevents testspecs and processes from being duplicated

junkerm:
-make TestSpecification and Process inherit from ISpecmateModelObject


junkerm:
@munterkalmsteiner definitely, that was my first attempt. However, changing the inheritance caused the UI to break at several places. Therefore I suggest to make this a PR of its own.

PR #365, opened 2018-12-28
fixes bug that causes the trace-link side-panel not to be shown when selecting a process step

junkerm:


junkerm:
@munterkalmsteiner In my understanding this is how it is supposed to be. According to specmate requirements, trace-links can (currenlty) only be set between  Process Steps and Requirements
https://trello.com/c/isYko1wb/274-anforderungen-mit-prozessschritten-verkn%C3%BCpfen

munterkalmsteiner:
Sorry, didn't know that!

PR #373, opened 2019-01-14
Added better naming rule for errors in CEG Nodes

Daedo:
Any validation errors of CEG Nodes will now be named with variable and condition instead of the node name.


junkerm:
In case of an isolated node, the variable/condition is printed twice, see screenshot. However it is only one unconnected node

image

junkerm (changes requested):
In case of an unconnected node (that is also a duplicate node) the node is listed twice in the error message that refers to the unconnected node. ( see screenshot)
image

junkerm:
I merged with the current develop to see if it's working with the latest ui chagnes. Now the error view isn't showing any errors anymore

image

junkerm:
@Daedo The branch doesn't build on travis
image

... and in fact it does not build locally on machine as well

junkerm (changes requested):
the typescript code is currently not building, neither in travis nor on my machine

Daedo:
It should run now.

junkerm (approved):
works!

PR #379, opened 2019-01-29
workaround for bug where deletetion of large models is not possible

junkerm:


bundles/specmate-persistency-cdo/src/com/specmate/persistency/cdo/internal/TransactionImpl.java
```diff
@@ -160,7 +160,10 @@ public boolean isDirty() {
 			comment.append(extractDeletedObjects(detachedObjects));
 		}
 
-		transaction.setCommitComment(comment.toString());
+		// FIXME: Workaround as CDO creates comment field with length 255
+		if(comment.length()<=255){
```
sebeder: Formatting is not right. Please fix.

PR #384, opened 2019-02-05
Major cleanup of data interface (and fixes)

sebeder:


junkerm (approved):
Tested -> test specifcatoins are shown
To check I also tried to generate a testspecification from a model and a model from the text. I also checked if the live-loading from Jira is affected by the changes
--> All works :-) 
--> Good Job!
--> Merge!

PR #385, opened 2019-02-05
Rewrite authentication method to handle 401 correctly

sebeder:


junkerm (approved):
Tested --> works
@eders: I added an additional small fix, could you re-review it and merge when you are fine.

PR #389, opened 2019-02-15
Fix/connection without target or source

sebeder:


sebeder:
Hi @junkerm , I just fixed the issue shown in the images. Are we done then?

PR #394, opened 2019-02-15
Fix csv export

junkerm:


bundles/specmate-testspecification/src/com/specmate/testspecification/internal/testskeleton/CSVTestSpecificationSkeleton.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.specmate.model.testspecification.ParameterAssignment;
+import com.specmate.model.testspecification.TestCase;
+import com.specmate.model.testspecification.TestParameter;
+import com.specmate.model.testspecification.TestSpecification;
+
+public class CSVTestSpecificationSkeleton extends BaseSkeleton {
+
+	private static final String TEXT_WRAP = "\"";
+	private static final String COL_SEP = ",";
```
sebeder: Set to ';', this way, Excel interprets it correctly right away.

PR #399, opened 2019-03-01
Fix/missing translations in ceg properties

sebeder:
I think something went wrong during a merge and thus, we need to do this again. The request is basically the same as last time.


junkerm (commented):
Can you additionally replace "Zustand" with "Bedingung" ?

junkerm (approved):

PR #407, opened 2019-03-14
remove formGroup and add ngModul

bliemelh:
Hi @junkerm the issue in edge, where checkboxes were wrongly checked should be repaired now


junkerm (approved):
looks good and works

PR #408, opened 2019-03-15
Adding null/undef-checks

sebeder:
https://trello.com/c/TXOtqcfb/366-l%C3%B6schen-einer-spezifikation-im-anforderungseditor-l%C3%A4%C3%9Ft-specmate-einfrieren-v-0215

https://trello.com/c/670dYE0W/372-liggesmeyer-beispiel-st%C3%BCrzt-ab

I could not reproduce this locally, which means something is wrong with the data on the dev env. However, this should fix the issues.


junkerm (approved):
looks good

PR #409, opened 2019-03-15
Removing valid check in save button

sebeder:
https://trello.com/c/uTRteWC6/373-fehlerhafte-modelle-m%C3%BCssen-gespeichert-werden-k%C3%B6nnen


junkerm (changes requested):
for certain validation error we must prohibit saving, e.g. when disallowed characters are entered into a name field, as the backend will reject these commits anyhow

image

sebeder:
Then, we need more clarification on when the button is disabled. I'll add a task for this.

PR #411, opened 2019-03-15
Catching all exceptions in connectors

sebeder:
This should fix the issue that the connector service crashes, when something bad happens in the connectors.


junkerm (approved):
looks good!

PR #412, opened 2019-03-15
enable external documentation directory

junkerm:


bundles/specmate-ui-core/src/com/specmate/ui/core/ResourceRegistration.java
```diff
@@ -1,27 +1,73 @@
...
+					try {
+						return new URL(docFolder.toURI().toURL(), name.substring(1));
+					} catch (MalformedURLException e) {
+						return null;
+					}
+				}
+
+				@Override
+				public String getMimeType(String name) {
+					return null;
```
sebeder, line 51: Really `null` as mime-type? Is this correct?

sebeder (changes requested):
Just a minor thing - If you can't fix the mime-type, just merge it.

bundles/specmate-ui-core/src/com/specmate/ui/core/ResourceRegistration.java
junkerm: null here just means: choose a mime-type
see https://osgi.org/javadoc/r5/cmpn/org/osgi/service/http/HttpContext.html#getMimeType(java.lang.String)

sebeder (approved):

sebeder (approved):

PR #414, opened 2019-03-19
Fix CSV Export

junkerm:
see comments here https://trello.com/c/bjJ5wuKV/349-csv-export-reihenfolge-von-spaltennamen-und-spalteninhalten-passt-nicht-zusammen-v-0215

correct sorting of input and output columns
add input/output bedore column name


bliemelh:
Hi @junkerm the sorting seems to work correctly now. 
Content with an "=" still appears as an equation, though. Maybe we have to escape the equal sign with a preceding ' sign.

PR #417, opened 2019-03-27
fix bad requirement names in connector task and add test

junkerm:
When a requirement name contains ",", "|", or ";" the validator will reject the commit. Hence we fix bad requirment names in the connector


junkerm:
@Daedo There were missing dependencies in the bnd file which caused the test to fail with classNotFound exception. It should work now, can you verify?

Daedo (approved):
Works and passes.

PR #421, opened 2019-04-03
Fix undo connection savecrash

wehrlean:
Fixed bug: Error, if the connection is undone. 
https://trello.com/c/zLhw1s0X


Daedo (changes requested):
Could you remove the /generated/ files and add the folders to .gitignore?
They probably shouldn't be in the repo.
Also you need to update the objects in the specmate dataservice otherwise the change to the object won't persist. If you add an edge, undo, safe, and then reopen the editor the crash will reappear.

wehrlean:
That's right, how could I remove it from an existing commit?

I don't think this is necessary, because you need to save, if you leave the editor and by saving the updated nodes will be persisted automatically. All other undo-actions are also only done on the cache.

Daedo:
Just use git -rm to delete the files in question. Other than that you're right.

PR #423, opened 2019-04-05
Enabling save button sometimes

sebeder:
This enables the save button on invalid models. But the save button is not enabled if:
- there is a string field that is longer than 4000 chars
- there is a name field that contains forbidden chars.

### DIFF
--- a/web/src/app/validation/text-length-validator.ts
+++ b/web/src/app/validation/text-length-validator.ts
@@ -1,0 +1,22 @@
+import { ElementValidatorBase } from './element-validator-base';
+import { IContainer } from '../model/IContainer';
+import { ValidationResult } from './validation-result';
+import { ValidationErrorSeverity } from './validation-error-severity';
+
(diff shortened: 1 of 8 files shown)


web/src/app/validation/text-length-validator.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ValidationResult } from './validation-result';
+import { ValidationErrorSeverity } from './validation-error-severity';
+
+export class TextLengthValidator extends ElementValidatorBase<IContainer> {
+    public validate(element: IContainer, contents?: IContainer[]): ValidationResult {
+        const keys = Object.keys(element);
+        for (let i = 0; i < keys.length; i++) {
+            const key = keys[i];
+            const currentAttribute = element[key];
+            if (typeof currentAttribute === 'string') {
```
junkerm, line 12: Can you make 'string' and '4000' a constant

junkerm (approved):
works! Great that you added the length>4000 check. I have just a minor comment, afterwards go ahead and merge

PR #428, opened 2019-04-11
Make text input resizable

bliemelh:
[Trello](https://trello.com/c/z6u5KCli/379-eingabefelder-der-testspezifikation-sind-mehrzeilig-und-in-der-gr%C3%B6%C3%9Fe-anpassbar)

I installed and used ngx-textarea-autosize to make the input fields in the testspecification resizable


sebeder (changes requested):
It's good to use a library for this. However, I think the dependency has to be included in the package.json file.

wehrlean (approved):
Looks good

PR #429, opened 2019-04-18
The method escapeString was failing when it got an empty string to check

bliemelh:
[trello](https://trello.com/c/Zu0HgGwu/391-fehler-bei-csv-export-einer-testspezifikiation-wenn-eine-wertzuweisung-nicht-belegt-ist)
Now escapeString is only called if the string is not empty.


bundles/specmate-testspecification/src/com/specmate/testspecification/internal/testskeleton/CSVTestSpecificationSkeleton.java
```diff
@@ -51,7 +51,8 @@ protected void generateTestCaseParameterAssignments(StringBuilder sb, List<Param
 		for (ParameterAssignment assignment : assignments) {
 			String assignmentValue = assignment.getCondition();
 			String characterToEscape = "=";
-			joiner.add(StringUtils.wrap(escapeString(assignmentValue, characterToEscape) + assignmentValue, TEXT_WRAP));
+			String escapeString = assignmentValue != null && !assignmentValue.isEmpty() ? escapeString(assignmentValue, characterToEscape) : "";
```
junkerm: You can use StringUtils.isEmpty, it checks for null and empty strings

junkerm (changes requested):
Could you add an test to our testsuite bundles/specmate-testspecification/test/com/specmate/testspecification/test/TestSkeletonTest.java
So far we only have a test for Java and not for CSV.

junkerm (approved):

PR #435, opened 2019-05-01
improve german pos and dependency analysis

junkerm:


Daedo (approved):
It doesn't look not good.

PR #439, opened 2019-05-07
Fix/unusual tree behaviour

tobi321:


bliemelh (approved):
Hi @tobi321 , everything seems to work!

PR #440, opened 2019-05-07
Fix/sort test cases

tobi321:


bliemelh (approved):
Hi @tobi321 , this looks good!

PR #442, opened 2019-05-09
Fixed typo

Daedo:
Major UI Overhaul


sebeder (approved):
That was too easy ;-)

PR #444, opened 2019-05-14
Ceg benchmark

wehrlean:
The benchmark for the CEG-model generation. It is devided into the three classes, one for english and two for german(with/without correct grammar within the nodes). 

https://trello.com/c/LlpaPOhA


bundles/specmate-integration-test/resources/DE/Dep_DE.spec
```diff
@@ -0,0 +1,226 @@
+def tagset dependencies DE.DEP.NEGRA {
```
junkerm: Is it necessary to have the rules in the test package? Then we have them redundantly at two places. If we change a rule in the model generation package and don't change it in the test package, the tests will not reflect the current state.

bundles/specmate-integration-test/src/com/specmate/test/integration/ModelGenerationTestBase.java
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	private String firstLetterLower(String string) {
+		return string.substring(0, 1).toLowerCase() + string.substring(1);
+	}
+	
+	//Check all combination of Nodes(lower/upper case, with/without condition...)
+	private boolean verifyConnections(CEGNode expSource, String srcId,  CEGNode expTarget, String targetId, JSONArray generated) {
+		if (expSource.getCondition().equals("")) {
+			return (EmfRestTestUtil.matches(generated,
```
junkerm: wouldn't it be easier to compare everything converted to lowercase?
Anyway, this method is way to long, can you pull out methods and give them a reasonalbe name?

junkerm (changes requested):
Please see my comments

junkerm (approved):
can you make the current set of tests pass and then merge this one and create a new PR as soon as you are done with the rest?

PR #447, opened 2019-05-15
Error cancel model-generation

wehrlean:
[Trello](https://trello.com/c/DKpAcAS2)

Cancel the model generation popup window, should now not lead to an exception any more. 
The throw clause was removed because if the promise is rejected (->cancel is pressed) nothing should happen.


bliemelh (approved):
Hi @wehrlean ! Works fine!

PR #448, opened 2019-05-20
Add the value property for the ecore model again

tobi321:
Reset Ecore model to previous state, since a mechanism for migration must first be added for a deleted attribute in order to delete the redundant 'value' property.


bliemelh:
Hi @tobi321 , this looks fine to me!

PR #450, opened 2019-05-27
Specmate Pattern Config

Daedo:
Specmate now has the option to specify the location of the CEG Pattern (*.spec) files in the config. Specmate back up to the local files, if no location is specified.


sebeder:
The Tests fail - And could you tell me how I can test this?

Daedo:
To test the functionality you have to add a path to a .spec file the specmate-config.properties file and see if the file is loaded when the loadRessources() function is called in the PatternbasedCEGGenerator.java class.

sebeder (approved):

PR #452, opened 2019-05-28
User statistic

tobi321:


Daedo:
From what I can see from my initial test, the gauge for currently logged in users does not work. So if I log in the metric page does indicate, that a login happend, but the number of logged in users is still zero.

tobi321:
This seems to be a cache related problem. If the username has still an entry in the database, the gauge  for currently logged in users is not incremented in order to avoid increments when a user performs multiple logins on different machines or browsers. 
Correct me if I am wrong, but from my point of view the observed behaviour should be correct.

Daedo (approved):

PR #456, opened 2019-06-11
Field change test procedure

tobi321:
The text input fields in the test procedure are also resizable.


sebeder (approved):
It's good! However, the tests failed. Could you please fix that. Then You're good to go.

PR #457, opened 2019-06-11
Fix projects not loaded

wehrlean:
[Trello](https://trello.com/c/MuyPYnZP)
Now the UI starts after the RestServlet.

Is the interface implemented in the correct way?


junkerm (changes requested):
The bundle specmate-ui-core ist not building on travis:
grafik
Reason: the bundle specmate-emfrest-api is not referenced in the build path of the bnd.bnd file. In the bnd file please replace the reference to specmate-emfrest with a reference to specmate-emfrest-api.
grafik

PR #459, opened 2019-06-14
Model generation input

Daedo:
There is now a textarea at the overview screen to input the CEG Model requirements.
If the user clicks at create model and the area is empty an empty CEG is generated.
Otherwise the CEG is filled by the automatically generated model. If no model could be generated the model is deleted.


sebeder (changes requested):
This is absolutely fine from a technical point of view. However, I think this should give an error message, if no model could be generated. Additionally, the data-service should be set to busy during the generation process (so that the small loading indicator on the top right spins).

sebeder (approved):
That's fine!

PR #461, opened 2019-06-17
Build docker image on travis

sebeder:
This does that specmate is pushed to the docker hub into the sebeder/specmate repository.


Dockerfile
```diff
@@ -1,37 +1,26 @@
 FROM ubuntu:18.04
```
junkerm, line 1: Is there a reason why you don't build upon the official adoptopenjdk docker image?

Dockerfile
```diff
@@ -1,37 +1,26 @@
...
-RUN npm run build-prod
 
-WORKDIR /opt/specmate/src/bundles
-RUN sed -i 's/\r$//' ./gradlew
-RUN sed -i 's/^M$//' ./gradlew
-RUN ./gradlew --version
-RUN ./gradlew --no-daemon clean build export -x check
-
-RUN cp specmate-std-env/generated/distributions/executable/dev-specmate-all.jar /opt/specmate/specmate.jar
+COPY bundles/specmate-std-env/generated/distributions/executable/dev-specmate-all.jar /opt/specmate/specmate.jar
```
junkerm: would it make semnse to use the prod version?

.travis.yml
```diff
@@ -3,34 +3,35 @@ language:
...
-- cd bundles/specmate-std-env/generated/distributions/executable/
-- mv dev-specmate-all.jar dev-specmate-all-$TRAVIS_TAG.jar 
-- mv prod-specmate-all.jar prod-specmate-all-$TRAVIS_TAG.jar
-- mv prod-specmate-cdo-server.jar prod-specmate-cdo-server-$TRAVIS_TAG.jar
-- mv prod-specmate-no-cdo-server.jar prod-specmate-no-cdo-server-$TRAVIS_TAG.jar
-- cd ../../../../..
+- (cd web && npm run build-prod)
+- (cd bundles && ./gradlew --no-daemon build --continue -x check)
+- (cd bundles && ./gradlew --no-daemon export -x check)
+# - (cd bundles/specmate-std-env/generated/distributions/executable/ && java -Xmx6000m -Djdk.crypto.KeyAgreement.legacyKDF=true -jar dev-specmate-all.jar &)
```
junkerm: remove the uncommented lines, or is there a reason you keep them?

junkerm (changes requested):
looks good, i have some questions and comments

Dockerfile
sebeder: It was easier to install the other dependencies on ubuntu that on the official adoptopenjdk image.

Dockerfile
sebeder: Would it? I don't have stakes in that. This was rather an arbitrary decision.

.travis.yml
sebeder: This disables the UI-Tests (because they were unstable). That's why we should keep them.

Dockerfile
junkerm: for the docker version I would use the prod version as it removes some unnecssary bundles

Dockerfile
junkerm: ok!

.travis.yml
junkerm: ok!

junkerm (approved):

PR #462, opened 2019-06-17
Objectif Monospaced Font

Daedo:
The requirement input field switches to a monospaced font, when the user types objectif code.

### DIFF
--- a/bundles/specmate-model/.project
+++ b/bundles/specmate-model/.project
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>specmate-model</name>
+	<comment>Project specmate-model created by Buildship.</comment>
+	<projects>
+	</projects>
+	<buildSpec>
+		<buildCommand>
+			<name>org.eclipse.buildship.core.gradleprojectbuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.buildship.core.gradleprojectnature</nature>
+	</natures>
+</projectDescription>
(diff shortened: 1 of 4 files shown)


bundles/specmate-model/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
junkerm, line 1: please remove this whole project

web/src/app/modules/views/main/editors/modules/contents-container/components/ceg-model-container.component.ts
```diff
@@ -31,21 +32,39 @@ export class CEGModelContainer extends TestSpecificationContentContainerBase<CEG
...
 
+    public isObjectif(): boolean {
+        let desc = this.modelDescription;
+
+        if (desc == undefined) {
+            desc = '';
+        }
+        desc = desc.trim();
+
+        if (desc.length <= 3) {
```
junkerm, line 45: Would it make sense to set this to 4?  This means as soon as someone types the whole word "WENN" the font is changed.
Also you could make the code much more simple by just doing description.startsWith(this.objectifIndicator)

web/src/app/modules/views/main/editors/modules/contents-container/components/ceg-model-container.component.ts
```diff
@@ -31,21 +32,39 @@ export class CEGModelContainer extends TestSpecificationContentContainerBase<CEG
...
+        let description = this.modelDescription;
+        if (description == undefined) {
+            description = '';
         }
-        this.modelDescription = this.modelDescription.trim();
+        description = description.trim();
 
-        if (this.modelDescription.length > 0) {
-            element.modelRequirements = this.modelDescription;
+        if (description.length > 0) {
```
junkerm, line 66: please comment on what is happening here or better, create a separate method with a speaking name

junkerm (changes requested):
Works! A few remarks below. We should think about what we do with the field in the eiditor view, Will it be removed or will we put the same functionality there too?

PR #463, opened 2019-06-21
check if project exists in authentication-process

bliemelh:
[Trello](https://trello.com/c/ltyCqAgI/418-nullpointerexception-wenn-project-nicht-richtig-%C3%BCbergeben-wird)

### DIFF
--- a/bundles/specmate-integration-test/src/com/specmate/test/integration/AuthenticationTest.java
+++ b/bundles/specmate-integration-test/src/com/specmate/test/integration/AuthenticationTest.java
@@ -63,15 +63,26 @@ public class AuthenticationTest extends EmfRestTest {
 	
 	@Test
 	public void testWrongProjectAtLogin() throws SpecmateException {
-		UserSession session = authenticationService.authenticate("resttest", "resttest", projectAName);
 		RestClient clientProjectA = new RestClient(REST_ENDPOINT, session.getId(), logService);
-			
-		String wrongProjectName = "wrongProjectName";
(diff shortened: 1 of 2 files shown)


junkerm (changes requested):
Looks good. Can you add Tests that check if Specmate aswers with a 500 if you try to Login with a wrong Project?

bundles/specmate-integration-test/src/com/specmate/test/integration/AuthenticationTest.java
```diff
@@ -1,85 +1,101 @@
...
+		assertEquals(ErrorCode.NO_AUTHORIZATION.getLiteral(), obj.get("ecode"));
+		assertTrue(((String) obj.get("detail")).contains(projectBName));
+		result.getResponse().close();
+
+		resetSelectedProject();
+	}
+	
+	@Test
+	public void testWrongProjectAtLogin() throws SpecmateException {
+		UserSession session = authenticationService.authenticate("resttest", "resttest", projectAName);
```
junkerm: do you need this line here? I guess it is not necessary for your test

bundles/specmate-integration-test/src/com/specmate/test/integration/AuthenticationTest.java
```diff
@@ -1,85 +1,101 @@
...
+		RestClient clientProjectA = new RestClient(REST_ENDPOINT, session.getId(), logService);
+			
+		String wrongProjectName = "wrongProjectName";
```
junkerm: I would also add a successfull login before and after the failed one to verify that the fault is indeed because of the wrong project name

junkerm (changes requested):
two suggestions for the test

PR #470, opened 2019-07-09
check for invalid characters in ceg node variable field

junkerm:
see https://trello.com/c/pwjm3rYv/452-probleme-mit-komma-im-titel-der-knoten-und-cegs


tobi321 (approved):
@junkerm looks good and works!

PR #473, opened 2019-08-02
Ceg benchmark german

wehrlean:
[Trello](https://trello.com/c/StHZO6oR)


bundles/specmate-integration-test/src/com/specmate/test/integration/ModelGenerationTestBase.java
```diff
@@ -31,19 +31,39 @@
...
+	}
+	
+	
+	/**
+	 * Check if two strings contain the same words. Order and case are not relevant
+	 * @param s1
+	 * @param s2
+	 * @return
+	 */
+	private boolean checkStringEquality(String s1, String s2) {
```
junkerm, line 52: This method seems overly complex due to the many splits and replaces of differnt kinds of whitspace. How about:
1) split strings with regex e.g. using \s* (any whitespace, any number of times)
2) Add all non empty parts into a set
3) Compare sets using equals

Additionally: I would opt to crate a class "TextUtit" in specmate-common and move the method there as a static method. I guess we can move more such methods there over time.

junkerm (changes requested):
- one of the model generation tests is failing
- a small comment regarding one of the utitlity functions
- please merge develop into this branch to check if all tests still run after that

wehrlean:
[Test](https://trello.com/c/OiEJ8TcF/1-test)

wehrlean:
[Test](https://trello.com/c/OiEJ8TcF/1-test)

PR #477, opened 2019-08-11
Fix bug test generation comparator

junkerm:
fixes bug in the test generator that occurs when two effects share the same variable


tobi321 (approved):
Works!

